PR #4, opened 2021-12-14
[BugFix] Fix compliance banner logic, reset recording stop and transcription stop case

lpx1989:
## Purpose
Previously have misleading message after starting and stopping both recording and transcription, then start transcription again, the resulting message is not very intuitive. And would need to dismiss the banner to reset state
impact: banner view text

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
Test recording/transcription banner will display correct message in teams interop meeting when starting/stopping recording/transcription from Teams client


iaulakh:
@lpx1989 I noticed that teams client indicate, Transcription stopped and UI library says Transcription saved. is it expected?
Screen Shot 2021-12-15 at 1 18 49 PM

lpx1989:
@iaulakh yea, we are not expecting to change the text for that. the main purpose for this bug fix is reset the state after both stops. and you can see on the iOS side, the text is same with us.

iaulakh (approved):

ShaunaSong (approved):

PR #17, opened 2022-01-06
[Refactor] drop context from meeting options

ahammer:
## Purpose

* Potential memory leak for API User. Storing Context inside a Data object we can't guarantee where it's retained can lead to a memory leak. Passing Context in via launch is considerably safer (ensures fresh/correct context, stays a param and never a field to be leaked).

```
[X] Yes
[ ] No
```

## Pull Request Type

2x MeetingOptions classes no longer store Context.
Context is passed via launch() when starting the composite.

For a user, they would need to add context to Launch. 

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Test the code

```
There should be no functional changes due to this code.
```

## What to Check
* You can still launch the composite


azure-communication-ui/azure-communication-ui-demo-app/src/main/java/com/azure/android/communication/ui/callingcompositedemoapp/launcher/CallingCompositeJavaLauncher.java
```diff
@@ -57,15 +57,15 @@ public void launch(final Context context,
 
         if (groupId != null) {
             final GroupCallOptions groupCallOptions =
-                    new GroupCallOptions(context, communicationTokenCredential, groupId, displayName);
```
ahammer, line 60: Since the normal use case is to create Options right before launch, the context really only needs to be moved down two lines (from construction of options, to the launch call)

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/GroupCallOptions.java
```diff
@@ -3,56 +3,43 @@
...
 
-import android.content.Context;
 import com.azure.android.communication.common.CommunicationTokenCredential;
 import java.util.UUID;
 
 /**
  * Options to start group call experience using {@link CallComposite}.
  */
 public final class GroupCallOptions {
-    private final Context context;
```
ahammer, line 14: Removed from these classes because it's not really an "Option" to the GroupCall. We also can not guarantee where an API user will store this Options class, but it's a risk for a context leak.

E.g. Say it is put in a redux store, or a global history repository. While safe to store data, not really safe to store a stale context, it can leak memory and fail if re-used.

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/TeamsMeetingOptions.java
```diff
@@ -3,55 +3,41 @@
...
 package com.azure.android.communication.ui;
 
-import android.content.Context;
 import com.azure.android.communication.common.CommunicationTokenCredential;
 
 /**
  * Options to start Teams meeting call experience using {@link CallComposite}.
  */
 public final class TeamsMeetingOptions {
-    private final Context context;
```
ahammer, line 13: same as above

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/CallComposite.java
```diff
@@ -60,13 +60,14 @@
      *
      * </pre>
      *
+     * @param context The android context used to start the Composite
```
pavelprystinka, line 63: Missing "." at the end

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/CallComposite.java
```diff
@@ -90,12 +91,15 @@ public void launch(final GroupCallOptions groupCallOptions) {
...
      * @param teamsMeetingOptions The {@link TeamsMeetingOptions} has parameters to
      *                            launch Teams meeting experience.
+     *
```
pavelprystinka, line 97: Not needed extra lines

pavelprystinka (changes requested):

iaulakh:
@ahammer can you please update changelog.md and github readme?

iaulakh:
tested, looks good.

lpeixin (dismissed):

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - API 21, 22 support
 
 ### Breaking Changes
+- Dropped Context from MeetingOptions, added to CallComposite.launch(..)
```
pavelprystinka, line 10: Lets have two separate records for these changes in this format:
-Remove Context from Class.Method(constructor)
-Add required parameter to Class.Method

pavelprystinka (changes requested):

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/CallComposite.java
ahammer: Installed a period

CHANGELOG.md
ahammer: Updated docs.

iaulakh (approved):

Mrayyan (approved):
👍

pavelprystinka (approved):

ShaunaSong (approved):

lpeixin (approved):

PR #41, opened 2022-01-27
[BugFix] Remove empty name field when user's name is blank/empty

Mrayyan:
## Purpose
Remove the name field background when user has no name or name is blank ("  ").

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Test the code

```
```

## What to Check
* User's name field background is hidden in participant grid when user has no name


azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/presentation/fragment/calling/participant/grid/ParticipantGridCellView.kt
```diff
@@ -65,6 +66,9 @@ internal class ParticipantGridCellView(
         val videoContainer: ConstraintLayout =
             findViewById(R.id.azure_communication_ui_participant_video_view_container)
 
+        val displayNameAndMicIndicatorViewContainer: LinearLayout =
```
ahammer, line 69: If you don't need anything out of View (I think you only use visibility here) it's better to just use View, one less import and the code is more robust to layout changes.

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/presentation/fragment/calling/participant/grid/cell/ParticipantGridCellVideoView.kt
```diff
@@ -102,14 +104,21 @@ internal class ParticipantGridCellVideoView(
...
+            displayNameOnVideoTextView.visibility = GONE
+        } else {
+            displayNameOnVideoTextView.text = displayName
+        }
     }
 
     private fun setMicButtonVisibility(isMicButtonVisible: Boolean) {
-
         if (!isMicButtonVisible) {
+            if (displayNameOnVideoTextView.visibility == GONE) {
```
ahammer, line 116: Is the visibility of another element the best way to express this condition?

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/presentation/fragment/calling/participant/grid/cell/ParticipantGridCellVideoView.kt
```diff
@@ -24,12 +25,13 @@ import kotlinx.coroutines.launch
 internal class ParticipantGridCellVideoView(
```
phileo99, line 25: do we care about seeing/using this View in Studio's Layout Editor?
If so, can we add the missing View constructors, if not can we add the annotation to suppress compiler warning

phileo99 (changes requested):

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/presentation/fragment/calling/participant/grid/cell/ParticipantGridCellVideoView.kt
Mrayyan: At the moment, yes, because the only way to "know" if the name is empty or not is to check if the name text view's visibility since we don't have access to displayName here. This is dependant on setDisplayName func where we set the text view's visibility to gone if the name is empty.

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/presentation/fragment/calling/participant/grid/ParticipantGridCellView.kt
Mrayyan: Sure. Changed in new commit.

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/presentation/fragment/calling/participant/grid/cell/ParticipantGridCellVideoView.kt
iaulakh: Muhammad, instead of having logic in view, can we create view model properties that will notify for visibility? We can also write test cases.

Mrayyan: @iaulakh Since that will require creating view model, I will just merge this PR for the bug fix and send another PR for that.

phileo99 (approved):
Tested
👍

ShaunaSong (approved):

PR #44, opened 2022-01-31
[Refactor] call end ordering

ahammer:
## Purpose

* There is a bug potentially in the call hang up (seen by others, but not reproduced). Investigation of the code hints at potential out of order hangup in respect to the Activity Lifecycle. 

This PR moves the activity finish() to the point after the call has successfully disconnected, instead of right when the dialog exit button is hit.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Test the code

```
```

## What to Check
* The you can hang up from meetings successfully


ahammer:
going to close for now so it doesn't hang. I'll re-open if needed or just re-apply.

PR #50, opened 2022-02-08
[Feature] process memory diagnostics

iaulakh:
## Purpose

* Added OMagnifier to enable memory diagnostics https://github.com/microsoft/OMagnifier 
* This is sample app feature, no code changes to UI library

Screen Shot 2022-02-08 at 11 02 40 AM
Screen Shot 2022-02-08 at 11 04 25 AM
Screen Shot 2022-02-08 at 11 04 50 AM

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
Enable memory diagnostics


azure-communication-ui/azure-communication-ui-demo-app/src/main/java/com/azure/android/communication/ui/callingcompositedemoapp/CallLauncherActivity.kt
```diff
@@ -182,20 +228,58 @@ class CallLauncherActivity : AppCompatActivity() {
...
-        runOnUiThread {
-            val builder = AlertDialog.Builder(this).apply {
-                setMessage(message)
-                setTitle("Alert")
-                setPositiveButton("OK") { _, _ ->
-                }
-            }
-            builder.show()
+    private fun saveState(outState: Bundle?) {
+        outState?.putBoolean(isTokenFunctionOptionSelected, callLauncherViewModel.isTokenFunctionOptionSelected)
```
ahammer, line 232: Instead of bundle, consider shared preferences so this will persist between sessions.

azure-communication-ui/azure-communication-ui-demo-app/src/main/java/com/azure/android/communication/ui/callingcompositedemoapp/CallLauncherActivity.kt
```diff
@@ -182,20 +228,58 @@ class CallLauncherActivity : AppCompatActivity() {
...
+        outState?.putBoolean(isKotlinLauncherOptionSelected, callLauncherViewModel.isKotlinLauncher)
+    }
+
+    private val onSampleListener = object : MemoryMonitor.OnSampleListener {
```
ahammer, line 236: Can put this object in another file, doesn't directly pertain to CallLauncherActivity

ahammer (changes requested):
I'd suggest considering decoupling from the Activity a bit more. (checkbox and state belong there though).

Since it's Application scoped, the farther from the activity the better, e.g. if you have API like.

MemoryViewer.initialize(application)
MemoryViewer.show()
MemoryViewer.hide()

The actual bits in the activity can be minimized?

Edit: Also, minimize the UI if you can. It only needs to be a really small on screen indicator, probably pinned in a corner.

azure-communication-ui/azure-communication-ui-demo-app/src/main/java/com/azure/android/communication/ui/callingcompositedemoapp/CallLauncherActivity.kt
iaulakh: I think we can modify this as part of separate Pr as this code is not touched in this PR.

azure-communication-ui/azure-communication-ui-demo-app/src/main/res/layout/memory_view.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
```
ahammer, line 1: I would suggest styling the Text with no background, small size. Make the text white and give it a heavy black shadow with offset (0,0), blur/spread of like 2.  Minimize padding and font size. That should make it highly readable while being out of the way and mostly translucent.

Maybe later/nice to have we can color text based on memory pressure?

ahammer (approved):

Mrayyan (approved):
Looks good. We would probably want to track more information about memory in the future (percentage of device's memory being used, etc.). This is a good tool in the sample app. I was looking into collecting some memory data for performance analysis. We will also have something like this in the UI library itself in the future.

PR #55, opened 2022-02-10
[Feature] fps diagnostics

iaulakh:
## Purpose

* FPSViewer is something a like alert. it can show the frame count every 1 second, developers can check if there is any blocker in main thread when they test. and it helped some developers to fix or avoid some heavy function on main thread.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/azure-communication-ui/azure-communication-ui-demo-app/src/main/java/com/azure/android/communication/ui/callingcompositedemoapp/diagnostics/FpsDiagnostics.kt
+++ b/azure-communication-ui/azure-communication-ui-demo-app/src/main/java/com/azure/android/communication/ui/callingcompositedemoapp/diagnostics/FpsDiagnostics.kt
@@ -12,12 +12,12 @@ class FpsDiagnostics private constructor(
 ) {
 
     companion object {
-        private var fpsDiagnostics: FpsDiagnostics? = null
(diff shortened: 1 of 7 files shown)


azure-communication-ui/azure-communication-ui-demo-app/src/main/java/com/azure/android/communication/ui/callingcompositedemoapp/diagnostics/FpsDiagnostics.kt
```diff
@@ -0,0 +1,37 @@
...
+import android.app.Application
+import com.microsoft.office.outlook.magnifierlib.Magnifier
+import com.microsoft.office.outlook.magnifierlib.frame.FPSMonitorConfig
+
+class FpsDiagnostics private constructor(
+    private val context: Application,
+) {
+
+    companion object {
+        private var fpsDiagnostics: FpsDiagnostics? = null
```
ahammer, line 15: can use lateinit and drop the ? I think.

ahammer (approved):

Mrayyan (approved):
Tested

PR #64, opened 2022-02-18
[Bug] Fix for crash when ending call if call isn't started

ahammer:
## Purpose

* There is a crash when ending a call before a join, this will resolve

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/redux/middleware/handler/CallingMiddlewareActionHandler.kt
+++ b/azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/redux/middleware/handler/CallingMiddlewareActionHandler.kt
@@ -127,6 +127,13 @@ internal class CallingMiddlewareActionHandlerImpl(
     }
 
     override fun endCall(store: Store<ReduxState>) {
+        val currentStatus = store.getCurrentState().callState.CallingStatus
+
+        // Do not end the call unless we are at least connecting state.
+        if (!listOf(CallingStatus.CONNECTED, CallingStatus.CONNECTING).contains(currentStatus)) {
+            return
+        }
+
         callingService.endCall()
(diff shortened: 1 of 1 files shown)


phileo99 (changes requested):
unable to join call on emulator, please take a look

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/redux/middleware/handler/CallingMiddlewareActionHandler.kt
```diff
@@ -127,6 +127,13 @@ internal class CallingMiddlewareActionHandlerImpl(
     }
 
     override fun endCall(store: Store<ReduxState>) {
+        val currentStatus = store.getCurrentState().callState.CallingStatus
```
iaulakh, line 130: can we move this logic to calling SDK wrapper? in case end call is called from somewhere else then we do not need to worry about crash.

iaulakh (changes requested):

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/redux/middleware/handler/CallingMiddlewareActionHandler.kt
ahammer: I tried originally, I'm going to close this and re-open after I investigate a bit more. May apply fix differently.

ahammer:
closing to investigate more, may re-open or apply a different fix. 

In my testing, this caused other bugs (e.g. I could not leave a call after joining it).

PR #69, opened 2022-02-23
[Bug] crash on early end

ahammer:
## Purpose

* Fix for crash on early back. Instead of not handling the exception (endCall called before start), we pass the error down, so it can bubble to the error handler. It now results in a prompt instead of a crash.

We might want to follow up and squash this error as well, but crash is the priority for this PR. We could return a CompletableFuture that does nothing instead of passing the error through, or we can follow up and ensure endCall() isn't called before startCall() so the condition for the error doesn't exist.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Test the code

```
Go to setup screen, and back. Do not join call. App should not crash.
```

## What to Check
* ...


ShaunaSong:
Tested.  App doesn't crash, showed alert instead.
Screen Shot 2022-02-22 at 8 56 33 PM

ShaunaSong (dismissed):

lpeixin (dismissed):

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/service/calling/sdk/CallingSDKWrapper.kt
```diff
@@ -82,6 +82,15 @@ internal class CallingSDKWrapper(
         callingSDKEventHandler.getRemoteParticipantInfoModelFlow()
 
     fun endCall(): CompletableFuture<Void> {
+        val call : Call?
+
+        // The getter for Call throws an exception, so we need to handle this
```
iaulakh, line 87: here, on back navigation, user get a message "call is not started". As, user want to navigate back, do we want to bubble this error? (User started the call, and want to end on back navigation, I think we do not need to display internal implementation message)

ahammer: It's quiet now.

iaulakh (approved):

phileo99 (approved):
tested, ran Espresso tests
LGTM
👍

PR #81, opened 2022-02-26
[Refactor] Ignore some UI elements and show up info header for user with Talkback on

lpeixin:
## Purpose
Ignore some UI elements for Talkback.
Modify the logic to always keep info header showing up when the Talkback is on

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

Test the App with Talkback on, use gesture(swipe left or right) to navigate the UI elements on Setup screen to verify if some UI elements have been ignored. (Also same for the participants' number on the info header).
Try to join the call with or without Talkback on, to see if the info header show up as expected.


ShaunaSong (approved):

ShaunaSong:
Tested.

Mrayyan (approved):

PR #86, opened 2022-03-01
[Feature] Reduce app size using Android R8

phileo99:
## Purpose

* Reduce application size using Android R8 code shrinking and obfuscation tool

```
[ ] Yes
[ X] No
```

## Pull Request Type

```
[ ] Bugfix
[ X] Feature
[ ] Other... Please describe:
```

## How to Test

* Test the code

```
build and run like normal
```

## What to Check
* can join a call


azure-communication-ui/azure-communication-ui-demo-app/build.gradle
```diff
@@ -90,12 +93,20 @@ dependencies {
...
+    }
     implementation 'androidx.activity:activity-ktx:1.3.1'
     implementation 'com.google.android.material:material:1.4.0'
     implementation "androidx.constraintlayout:constraintlayout:$constraint_layout"
     implementation 'com.github.kittinunf.fuel:fuel:2.3.1'
-    implementation "com.azure.android:azure-communication-calling:$calling_sdk_version"
+    implementation "com.azure.android:azure-core:$azure_core_version"
+    implementation "com.google.code.gson:gson:$gson_version"
+    implementation ("com.azure.android:azure-communication-calling:$calling_sdk_version") {
+        exclude group: "com.azure.android", module: "azure-core"
```
ahammer, line 106: Are these excludes followed by re-imports necessary? E.g. GSON.

ahammer: Also, risk of breakages?

ahammer: Also, side question? Why do we bring azure-communication-calling into demo, shouldn't it come transitively via azure-communication-ui?

phileo99: The gson version update is a dependency that CallingSDK depends on.  The reason I wanted to update to newest GSON is to fix a [security warning](https://skype.visualstudio.com/SCC/_componentGovernance/client_calling_spool-native/alert/6140890?typeId=5727151) about the older version of GSON.  Technically it's not our responsibility but it looked like low hanging fruit.  I can remove if you want.

we bring azure-communication-calling into demo because the Error Handler is a CallingSDK construct

iaulakh:
@phileo99 In my opinion, we do not need to apply size reduction to our demo application. Our goal was to write document to share how the size can be reduced. 

Applying size reductions will cause any issue in future? (like, skipping few files on SDK update)

azure-communication-ui/azure-communication-ui-demo-app/build.gradle
pavelprystinka: I think we should not make exclusion of what CallingSDK team decided to include. It should be up to them to make this decision.

pavelprystinka (dismissed):

azure-communication-ui/azure-communication-ui-demo-app/build.gradle
phileo99: @pavelprystinka azure-communication-ui library module line 84 is already including newer version of azure core lib, should that be removed?

phileo99: @pavelprystinka  ok nevermind, I found a solution

phileo99:
> @phileo99 In my opinion, we do not need to apply size reduction to our demo application. Our goal was to write document to share how the size can be reduced.
> 
> Applying size reductions will cause any issue in future? (like, skipping few files on SDK update)

@iaulakh most, if not all of our customers that publishes to Google Play store are doing so with code obfuscation enabled (either Proguard or R8).  Therefore we need to support code obfuscation as part of the app size reduction process.
Most of the heavy lifting to setup support for R8 has been done in this PR, any issues in future will be surfaced in bug bashes.
acs-ui-library.pro is inter-operable with both Proguard or R8

ahammer (dismissed):
Make sure
-dontobfuscate

flag is in our R8, being open source we have no imperative to obfuscate our builds.

phileo99:
@ahammer done, I've added `-dontobfuscate` to acs-ui-library.pro

Mohtasim (dismissed):

ShaunaSong (dismissed):

phileo99:
P4a Without Proguard
apk-no-proguard-Screenshot_20220329-231724

P4a With Proguard
apk-with-proguard-Screenshot_20220329-232719

azure-communication-ui/azure-communication-ui-demo-app/acs-ui-library.pro
```diff
@@ -0,0 +1,92 @@
+# Add project specific ProGuard rules here.
```
ahammer, line 1: Probably don't need the built in comments in the files.

ahammer (dismissed):

ahammer (approved):

Mrayyan (approved):
👍

ShaunaSong (approved):

Mohtasim (approved):

PR #102, opened 2022-03-09
[Feature] Let Talkback read confirm leave overlay properly

lpeixin:
## Purpose
When confirm leave overlay is up, Talkback should focus on the buttons on the overlay. 
When confirm leave overlay is up and swiping left or right to navigate the UI elements, Talkback should not focus on the UI elements underneath.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

Turn on screen reader(eg. talkback on pixel)
run the app to join group call with remote participants
using swipe to focus on leave call button on the bottom bar
double click to open confirm leave overlay
verify: 1. the first element being focused is blue button on the overlay.
            2. swiping left or right, screen reader should not focus on any UI elements under confirm leave overlay


iaulakh:
@lpx1989 can we sync with iOS team if this state is added?

ShaunaSong (approved):
Tested

PR #145, opened 2022-03-28
[BUG] Audio Issues when Rotating Device

ahammer:
## Purpose

- Bluetooth was cutting out
- Speaker was jumping back to Bluetooth

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/presentation/manager/AudioSessionManager.kt
+++ b/azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/presentation/manager/AudioSessionManager.kt
@@ -28,9 +28,9 @@ import com.azure.android.communication.ui.redux.state.PermissionStatus
 internal class AudioSessionManager(
     private val store: Store<ReduxState>,
     private val context: Context,
-    private val audioManager: AudioManager,
 ) : BluetoothProfile.ServiceListener, BroadcastReceiver() {
 
+    private val audioManager by lazy { context.getSystemService(Context.AUDIO_SERVICE) as AudioManager }
     private var bluetoothAudioProxy: BluetoothHeadset? = null
 
     private val isBluetoothScoAvailable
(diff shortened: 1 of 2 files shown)


ShaunaSong (dismissed):

jamchengms (commented):
one minor thing i noticed in code

val audioManager = context.getSystemService(Context.AUDIO_SERVICE) as AudioManager
audioManager is initialized in isHeadsetActive() instead of start

does it make sense we only set the audioManager once in the start() instead?

ahammer:
> one minor thing i noticed in code
> 
> val audioManager = context.getSystemService(Context.AUDIO_SERVICE) as AudioManager audioManager is initialized in isHeadsetActive() instead of start
> 
> does it make sense we only set the audioManager once in the start() instead?

Deleted, it comes from constructor, didn't need it there.

jamchengms (dismissed):

lpeixin (dismissed):

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/presentation/manager/AudioSessionManager.kt
```diff
@@ -61,19 +62,23 @@ internal class AudioSessionManager(
 
     suspend fun start() {
 
+        // On first launch we need to init the redux-state, check Bluetooth and Headset status
+        if (!initialized) {
+            initializeAudioDeviceState()
+            updateBluetoothStatus()
```
phileo99, line 68: is `updateBluetoothStatus()` guaranteed to be called after `onServiceConnected()`?

ahammer: Yeah maybe, I'll look. In testing it didn't switch on rotation, but might be undesirable.

ahammer: @phileo99  Actually, I might do this differently.

Going to retest. UpdateBluetoothStatus needs to be called at different times, e.g. after service, or after permissions granted. (first launch vs other launches).  

I'm going to change the fix and test again.

iaulakh (approved):

jamchengms (approved):

phileo99 (approved):
tested, ran Espresso UI tests.
LGTM
👍

PR #197, opened 2022-04-13
[Feature] Detect OS Locale and Refactor Public API

ShaunaSong:
## Purpose

* Refactor the API, and let Contoso pass Locale Object
* Rename LanguageCode into SupportLanguage
* Detect the OS Locale as Default 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test

* Open Setting Page
* Click on Language DropDown 
* Language Selection still working 

## What to Check
*  Did not break language selection


azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/configuration/LocalizationConfiguration.java
```diff
@@ -26,38 +28,36 @@
  * @see CallComposite
  */
 public final class LocalizationConfiguration {
-    private final LanguageCode languageCode;
     private int layoutDirection;
```
Mohtasim, line 31: Can we make this variable final as well?

ShaunaSong: Because we don't need it in the every constructor so that we cannot make it final

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/configuration/AppLocalizationProvider.kt
```diff
@@ -3,14 +3,11 @@
...
 package com.azure.android.communication.ui.configuration
 
-internal class AppLocalizationProvider : LocalizationProvider {
-    private lateinit var language: String
-    private var customTranslation: Map<String, String>? = null
+import java.util.Locale
 
+internal class AppLocalizationProvider : LocalizationProvider {
+    private lateinit var locale: Locale
     override fun apply(localeConfig: LocalizationConfiguration) {
```
phileo99, line 10: I can't seem to find any usage of this in the project - where is this being used?

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/presentation/CallCompositeActivity.kt
```diff
@@ -146,19 +143,14 @@ internal class CallCompositeActivity : AppCompatActivity() {
...
-            val languageAttributes = localeConfig.languageCode.toString().split("-")
-            val languageCode = languageAttributes[0]
-            val countryCode =
-                if (languageAttributes.size > 1) languageAttributes[1] else ""
-            config.setLocale(Locale(languageCode, countryCode))
-            resources.updateConfiguration(config, resources.displayMetrics)
-            localizationProvider.apply(localeConfig)
+            config.setLocale(localeConfig.locale)
         }
+        resources.updateConfiguration(config, resources.displayMetrics)
```
phileo99, line 153: This is deprecated in API 25, please take a look at how to handle it for API 25 and newer

phileo99 (changes requested):

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/presentation/CallCompositeActivity.kt
ShaunaSong: Yea, there is no easy and new native way to replace this method.  Will leave it for now. 

https://stackoverflow.com/questions/39705739/android-n-change-language-programmatically/40849142#40849142

phileo99 (approved):
Tested.  The deprecated function call will be addressed in a separate ADO ticket
👍

Mohtasim (approved):

PR #199, opened 2022-04-14
[Feature] Handle call eviction

phileo99:
## Purpose

* Re-use existing Error Handling flow to also handle call eviction

```
[ ] Yes
[ X] No
```

## Pull Request Type

```
[ ] Bugfix
[ X] Feature
[ ] Other... Please describe:
```

## How to Test

* Join Group or Teams call
* Using Teams client or browser for group call, Remove yourself from the call

## What to Check
* verify that you are back in Setup screen
* verify that there is a snackbar
* verify that the Snackbar displays a message about call eviction


azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/service/calling/CallingService.kt
```diff
@@ -43,6 +44,12 @@ internal class CallingService(
         * Expected behavior.
         * */
         private const val CALL_END_REASON_DECLINED = 603
+        internal const val CALL_END_REASON_EVICTED = 5000
```
ahammer, line 47: any reason to diverge here with internal instead of private? 

Since it's a library of error codes, I think making them all internal is OK here for consistency.

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/service/calling/CallingService.kt
```diff
@@ -122,7 +129,14 @@ internal class CallingService(
         coroutineScope.launch {
             callingSDKWrapper.getCallingStateWrapperSharedFlow().collect {
                 val callStateError = when (it.callEndReason) {
-                    CALL_END_REASON_SUCCESS, CALL_END_REASON_CANCELED, CALL_END_REASON_DECLINED -> null
+                    CALL_END_REASON_SUCCESS -> {
```
ahammer, line 132: I feel with kotlin this would be inline if would read better.

CALL_END_REASON_SUCCESS -> if (isEvicted(it)) CallStateError(...) else null

something about an else block with just "null" in it, or leaving out the returns just feels off here.

azure-communication-ui/azure-communication-ui/src/main/res/values-ja/azure_communication_ui_strings.xml
```diff
@@ -71,7 +71,7 @@
     <string name="azure_communication_ui_snack_bar_button_dismiss">却下</string>
     <string name="azure_communication_ui_cal_state_error_call_end">エラーが発生したため、通話から削除されました。</string>
     <string name="azure_communication_ui_snack_bar_text_error_call_join">エラーが発生したため、通話に参加できません。</string>
-
+    <string name="azure_communication_ui_call_state_call_evicted"></string>
```
Mohtasim, line 74: Instead of using call twice, could we use 'azure_communication_ui_call_state_evicted'?

azure-communication-ui/azure-communication-ui/src/main/res/values-ja/azure_communication_ui_strings.xml
```diff
@@ -71,7 +71,7 @@
     <string name="azure_communication_ui_snack_bar_button_dismiss">却下</string>
     <string name="azure_communication_ui_cal_state_error_call_end">エラーが発生したため、通話から削除されました。</string>
```
Mohtasim, line 72: Hi, just saw this typo. But we should refactor the spelling of call here.

azure-communication-ui/azure-communication-ui/src/main/res/values-ja/azure_communication_ui_strings.xml
Mohtasim, line 74: Also call eviction string is empty? Just checking if this is our intended action. According to 'what to check' section, there was suppose to be snacker with some string?

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/service/calling/sdk/CallingStateWrapper.kt
```diff
@@ -5,4 +5,8 @@ package com.azure.android.communication.ui.service.calling.sdk
 
 import com.azure.android.communication.calling.CallState
 
-internal data class CallingStateWrapper(val callState: CallState, val callEndReason: Int)
+internal data class CallingStateWrapper(
+    val callState: CallState,
+    val callEndReason: Int,
+    val callEndReasonExtra: Int = 0
```
ahammer, line 11: Would recommend "Code" instead of "Extra" here to maintain cohesion with the Calling SDK

ahammer (dismissed):
Have a look at feedback

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/service/calling/CallingService.kt
phileo99: I needed it exposed as internal for writing the Unit Tests on call eviction.  Which kinda says we don't have Unit tests for  CALL_END_REASON_TOKEN_EXPIRED, CALL_END_REASON_DECLINED, CALL_END_REASON_CANCELED.  I'll change the other constants to internal for consistency.
@Mohtasim can you write up child tickets for the Unit test user story that you created ?

azure-communication-ui/azure-communication-ui/src/main/res/values-ja/azure_communication_ui_strings.xml
phileo99: @Mohtasim based on input from @ShaunaSong I have left the translations empty for now, it will be taken care of by the Touchdown service which should auto generate translations

azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/service/calling/CallingService.kt
phileo99: @ahammer I have moved this into getCallStateError() function

azure-communication-ui/azure-communication-ui/src/main/res/values-ja/azure_communication_ui_strings.xml
phileo99: fixed

Mohtasim (approved):
Looks good to me

Mrayyan (approved):
👍

PR #220, opened 2022-04-27
[Bug] Fix crash with removal of camera permission while call ongoing

dhiraj:
## Purpose

* Fixes a crash that occurs if a user deliberately denies video camera permission while a call is ongoing by backgrounding the app and going to App Info -> Permissions.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Open the app and join a call
* Turn on video camera, providing permission if required
* Background the app using circle button and open up App Info using long press on our app icon
* Navigate to Permissions and turn off Camera permission
* Switch back to the (now no longer running demo app)
* The app should finish the directly launched `CallCompositeActivity` _silently_
* Re-joining the call should work and you should get a disabled Camera button

## What to Check
* Try turning off-and-on permissions for Camera and Microphone while a call is ongoing
* The demo app should behave reliably for permission changes, without crashing and also not cause any kind of loop between Activity starting and finishing early

## Other Information
* Added a new function `hasConfig(id:Int):Boolean` to `CallCompositeConfiguration` that allows us to check if a given instanceId is valid without causing an exception to get thrown
* `DependencyInjectionContainerHolder` now checks whether any assigned `instanceId` is present in configs and throws an `IllegalArgumentException` if it isn't found
* `CallCompositeActivity` will call `finish()` and do an early return during `onCreate()` if setting `instanceId` fails
* Checks had to be added in `onDestroy()` of `CallCompositeActivity`, `CallingFragment` and `SetupFragment` because we would still try to deallocate resources that were now not being assigned in this situation


azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/presentation/CallCompositeActivity.kt
```diff
@@ -50,6 +51,10 @@ internal class CallCompositeActivity : AppCompatActivity() {
     private val instanceId get() = intent.getIntExtra(KEY_INSTANCE_ID, -1)
 
     override fun onDestroy() {
+        if (!CallCompositeConfiguration.hasConfig(instanceId)) {
```
ahammer, line 54: Leave a reasoning here. It's a small enough block without enough context that the "why" will get easily lost.

e.g. Covers edge case where Android tries to recreate call activity after process death (e.g. due to revoked permission). If no configs are detected we can just exit without cleanup.

dhiraj: Resolved with 21f4ef3

phileo99 (dismissed):
tested, looks good
👍

iaulakh (changes requested):
Hi @dhiraj, thank you for fixing complex long standing bug. I noticed that the branch has memory leak. It will be great to investigate the cause of leaks.

steps:
Start a fresh app (new install)
Accept audio permissions and join call
During call try to turn on camera and reject video permissions
End call and start call again
Open Android Studio memory profiler and profile memory
Few memory leaks are present
Screen Shot 2022-04-30 at 12 32 51 AM

dhiraj:
@iaulakh I've fixed the memory leak that you found; I couldn't reproduce easily using Android Studio, but Leak Canary was directly showing the same leak to me on the device which has now stopped happening with the changes in commit 098b6e5. I've chosen to always check that the `lateinit` vars actually have been initialized before calling the `stop()` methods on them, this way they will be checked irrespective of whether they were inited as part of normal lifecycle or were left uninitialized as part of granted permission later being denied causing an early `finish()` of the `CallCompositeActivity`.

Thanks for catching this memory leak, I'd appreciate it if you could profile again and confirm that it does not leak anymore.

iaulakh (approved):
tested, looks good

phileo99 (approved):
tested, looks good
👍

PR #222, opened 2022-04-28
[Bug] Fix control bar spacing for phone to match Figma

dhiraj:
## Purpose

* The layout was packing the control bar buttons together, whereas the Figma screen design has them spread out.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Use a phone and a tablet device to test control bar button layout when in call

## What to Check
* Buttons should be spread out on phone devices in both landscape and portrait orientations, but should continue to be packed together for the tablet layout

## Other Information
* I've refactored the layouts to now use a `ConstraintLayout` with the buttons in a vertical or horizontal chain so that we can easily control the spacing flow using chain styles
* On tablet there is no difference in the layout for both landscape and portrait orientations (but the buttons have the `packed` chainStyle) so I've created a new layout file for tablet devices, and then used `<include>` to use the same layout for both orientations

### Phone Portrait
Screenshot_1651175537

### Phone Landscape Left
Screenshot_1651175597

### Phone Landscape Right
Screenshot_1651175447

### Tablet Portrait
Screenshot_1651170971

### Tablet Landscape
Screenshot_1651170975


Mrayyan (approved):
Looks good to me 👍

phileo99 (approved):
tested call screen in portrait and landscape, looks good
👍

PR #301, opened 2022-06-01
[Bug] AudioFocus memory leak

dhiraj:
## Purpose

* Fixes leak at onFocusChange listener function var

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Open the app from Android Studio in memory profiler mode and start a call
* End the call
* Repeat several times (3-5)
* End the Activity (use hardware back button to get out of the app)
* Test mic-hold-release as per call hold feature tests (should auto-hold when another call starts, shouldn't allow resuming call when mic is not available, should auto-resume call when possible)

## What to Check
* Without this PR the listener function is retained and the AudioFocusManager is being held in memory

### DIFF
--- a/azure-communication-ui/azure-communication-ui/src/calling/java/com/azure/android/communication/ui/calling/presentation/manager/AudioFocusManager.kt
+++ b/azure-communication-ui/azure-communication-ui/src/calling/java/com/azure/android/communication/ui/calling/presentation/manager/AudioFocusManager.kt
@@ -31,18 +31,18 @@ internal abstract class AudioFocusHandler : AudioManager.OnAudioFocusChangeListe
 // Newer API Version of AudioFocusHandler
 @RequiresApi(Build.VERSION_CODES.O)
 internal class AudioFocusHandler26(val context: Context) : AudioFocusHandler() {
-    private val audioManager = context.getSystemService(Context.AUDIO_SERVICE) as AudioManager
+    private fun audioManager() = context.getSystemService(Context.AUDIO_SERVICE) as AudioManager
 
     private val audioFocusRequest26 = AudioFocusRequest.Builder(AudioManager.AUDIOFOCUS_GAIN)
         .setOnAudioFocusChangeListener(this).build()
 
     override fun getAudioFocus() =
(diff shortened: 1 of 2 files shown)


Mohtasim (dismissed):
Looks good to me

azure-communication-ui/azure-communication-ui/src/calling/java/com/azure/android/communication/ui/calling/presentation/manager/AudioFocusManager.kt
```diff
@@ -31,34 +31,34 @@ internal abstract class AudioFocusHandler : AudioManager.OnAudioFocusChangeListe
...
     override fun releaseAudioFocus() =
-        audioManager.abandonAudioFocusRequest(audioFocusRequest26) == AudioManager.AUDIOFOCUS_REQUEST_GRANTED
+        audioManager().abandonAudioFocusRequest(audioFocusRequest26) == AudioManager.AUDIOFOCUS_REQUEST_GRANTED
 }
 
 // Legacy AudioFocus API
 @Suppress("DEPRECATION")
 internal class AudioFocusHandlerLegacy(val context: Context) : AudioFocusHandler() {
-    private val audioManager = context.getSystemService(Context.AUDIO_SERVICE) as AudioManager
+    private fun audioManager() = context.getSystemService(Context.AUDIO_SERVICE) as AudioManager
```
iaulakh, line 51: I think we can do same here  private fun audioManager()?

iaulakh (approved):

azure-communication-ui/azure-communication-ui/src/calling/java/com/azure/android/communication/ui/calling/presentation/manager/AudioFocusManager.kt
dhiraj: Done.

Mohtasim (approved):
Tested and Looks good to me.

PR #308, opened 2022-06-02
[Bug] fix call hold not triggering on some devices

dhiraj:
## Purpose

* There is a possibility that earlier change from making the audioManager a function call instead of the instance variable causes some devices to not be able to receive call hold status change events
* This PR reverts the function change, so that we can test if the revert fixes the problem

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Open the app
* Start a call
* While in the ACS call, receive a parallel call from Whatsapp, or PSTN etc.

## What to Check
* Verify that all call-on-hold requirements are being met on your device


azure-communication-ui/azure-communication-ui/src/calling/java/com/azure/android/communication/ui/calling/presentation/manager/AudioFocusManager.kt
```diff
@@ -31,34 +31,34 @@ internal abstract class AudioFocusHandler : AudioManager.OnAudioFocusChangeListe
 // Newer API Version of AudioFocusHandler
 @RequiresApi(Build.VERSION_CODES.O)
 internal class AudioFocusHandler26(val context: Context) : AudioFocusHandler() {
```
phileo99, line 33: both this class and `AudioFocusHandlerLegacy` do not depend on Context, can we pass in an immutable instance of AudioManager instead of Context

PR #309, opened 2022-06-02
[Bug] fix call hold crash

dhiraj:
## Purpose

* Fixes a crash that would be reproduced by quitting the call while the call hold resume button overlay was showing

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Open the app, start an ACS call
* Receive another call from another service like PSTN / Whatsapp etc.
* Come back to the ACS sample app; our resume call overlay should be showing
* Quit the call by hardware back button

## What to Check
* App should no longer crash in this situation


iaulakh (approved):

phileo99 (approved):
👍

Mrayyan (approved):

PR #342, opened 2022-06-14
Refactor/kotlin redux

ahammer:
## Purpose

* Migrate to Kotlin-Redux from Bespoke

```
- [ ] Yes
- [ ] No
```

## Pull Request Type

```
- [ ] Bugfix
- [ ] Feature
```
## Pull Request Checklist

```
- [ ] Public API changes
- [ ] UI Changes
- [ ] Tests
- [ ] Unit Tests Included
```

## How to Test

* Open...
* Click on...

## What to Check
* ...


ahammer:
Arguments for: Defer some of our complexity to an external. Re-use of redux pattern across libraries without copy/paste. Library version has a cleaner more robust/implementation (e.g. coroutines not necessary, no handler thread we need to manager, thread safety guaranteed). Redux is very standard and unlikely to change API's. Since Kotlin multiplatform support, possibility to have one Redux state/actions across both Android/iOS. Ability to integrate standardized middleware.

Arguments Against: Adding a library means a transitive dependency. If there are bugs, they are harder for us to change on (PR/Publishing of an external). While redux is not commonly used, we'd have to think about the transitive dependency and if it's something that might conflict downstream.

Alternatives: Other redux libraries? Shadowing/Forking (discouraged). Audit and upgrade our internal library? Decouple and publish our own redux framework so we can re-use and share with community (would increase engagement and visibility of our libraries).

iaulakh:
Hi Adam, previously, we had discussion to use external library and decided not to use because of following reasons:

Using a library for redux or any other library can became an issue for following scenario:
If Contoso is using same library and different version then integrating UI library will become issue. Either we need to match version or Contoso need to match version. 

If Contoso updated to higher version, it may bring issues to UI library too.

ahammer:
Going to drop this one, I don't want to follow this path as the risks of a 3rd party are too great

ahammer:
@iaulakh 

I think what might make more sense is for us to break out our own arch components for now into a library (Service Locator + Redux), and use for Chat. I have in another branch.

As chat approaches public preview, we can decide to either a) publish the arch library (and potentially upgrade call to it), or b) merge arch library into the chat library.

I don't want to refactor anything regarding redux (besides middleware where some stuff can be improved) in Calling for now, unless there is a good use case, but as we approach CallWithChat, it'd be nice to have the option of a single Service Locator/Redux instance and implementation as opposed to 3.

iaulakh:
@ahammer I agree. for chat, we can extract redux layer and in future use same for call, call +chat.

I think for timing we can start with chat, in separate module and can extract to library if required in future.

PR #343, opened 2022-06-15
[Feature] multiple event handlers

pavelprystinka:
## Purpose

* Allow multiple event handlers

- [X] Yes
- [ ] No

## Pull Request Type

- [x] Bugfix
- [X] Feature

## Pull Request Checklist

- [X] Public API changes
- [ ] UI Changes
- [ ] Tests
- [ ] Unit Tests Included

## How to Test

* Open...
* Click on...

## What to Check
* ...

### DIFF
--- a/azure-communication-ui/azure-communication-ui/src/calling/java/com/azure/android/communication/ui/calling/configuration/events/CallCompositeEventsHandler.kt
+++ b/azure-communication-ui/azure-communication-ui/src/calling/java/com/azure/android/communication/ui/calling/configuration/events/CallCompositeEventsHandler.kt
@@ -22,9 +22,9 @@ internal class CallCompositeEventsHandler {
 
     fun getOnRemoteParticipantJoinedHandlers() = remoteParticipantJoinedHandler.asIterable()
 
-    fun addOnRemoteParticipantJoinedHandler(handler: CallCompositeEventHandler<CallCompositeRemoteParticipantJoinedEvent>) =
+    fun addOnRemoteParticipantJoinedEventHandler(handler: CallCompositeEventHandler<CallCompositeRemoteParticipantJoinedEvent>) =
         remoteParticipantJoinedHandler.add(handler)
 
-    fun removeOnRemoteParticipantJoinedHandler(handler: CallCompositeEventHandler<CallCompositeRemoteParticipantJoinedEvent>) =
+    fun removeOnRemoteParticipantJoinedEventHandler(handler: CallCompositeEventHandler<CallCompositeRemoteParticipantJoinedEvent>) =
         remoteParticipantJoinedHandler.remove(handler)
 }
--- a/azure-communication-ui/azure-communication-ui/src/calling/java/com/azure/android/communication/ui/calling/configuration/events/CallCompositeEventsHandler.kt
+++ b/azure-communication-ui/azure-communication-ui/src/calling/java/com/azure/android/communication/ui/calling/configuration/events/CallCompositeEventsHandler.kt
@@ -10,7 +10,7 @@ import com.azure.android.communication.ui.calling.models.CallCompositeRemotePart
 internal class CallCompositeEventsHandler {
     private val errorHandlers = mutableSetOf<CallCompositeEventHandler<CallCompositeErrorEvent>>()
     private val remoteParticipantJoinedHandler =
-            mutableSetOf<CallCompositeEventHandler<CallCompositeRemoteParticipantJoinedEvent>>()
+        mutableSetOf<CallCompositeEventHandler<CallCompositeRemoteParticipantJoinedEvent>>()
 
     fun getOnErrorHandlers() = errorHandlers.asIterable()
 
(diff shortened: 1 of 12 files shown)


azure-communication-ui/azure-communication-ui/src/calling/java/com/azure/android/communication/ui/calling/configuration/events/CallCompositeEventsHandler.kt
```diff
@@ -8,19 +8,23 @@ import com.azure.android.communication.ui.calling.models.CallCompositeErrorEvent
 import com.azure.android.communication.ui.calling.models.CallCompositeRemoteParticipantJoinedEvent
 
 internal class CallCompositeEventsHandler {
-    private var errorHandlers: CallCompositeEventHandler<CallCompositeErrorEvent>? = null
-    private var remoteParticipantJoinedHandler: CallCompositeEventHandler<CallCompositeRemoteParticipantJoinedEvent>? =
-        null
+    private val errorHandlers = mutableSetOf<CallCompositeEventHandler<CallCompositeErrorEvent>>()
```
grigoryk, line 11: `add`/`remove` methods may be called from different threads, right? Then access to `errorHandlers` (and `remoteParticipantJoinedHandler`) should be synchronized.

Also, I wonder if a Set here is a good choice - it's unordered, and applications may assume that adding handler1 before handler2 means that they will be called in that order, which isn't guaranteed if you use a set. But it does protect against accidentally adding the same thing multiple times, so perhaps that's all right.

azure-communication-ui/azure-communication-ui/src/calling/java/com/azure/android/communication/ui/calling/configuration/events/CallCompositeEventsHandler.kt
```diff
@@ -8,19 +8,23 @@ import com.azure.android.communication.ui.calling.models.CallCompositeErrorEvent
...
+    private val remoteParticipantJoinedHandler =
```
grigoryk, line 12: For future, as there are more of these handler types, it may be worth it to generalize this a tiny bit into a "handler registry" class of sorts? Something defined against `handlers: mutableSetOf<T>`, with `add`/`remove`/`notify` methods and proper synchronization.

docs/CHANGELOG_UI_CALLING.md
```diff
@@ -1,6 +1,14 @@
-# Release History
+# Calling library Release History
 
-## 1.0.0-beta.2 (2022-06-13)
+
+## Calling 1.0.0.0 (FUTURE)
+
+### Breaking Changes
+- `CallComposite.setOnErrorHandler` is replaced with `CallComposite.addOnErrorHandler`
```
grigoryk, line 7: nit: the actual method name is `addOnErrorEventHandler`.

azure-communication-ui/azure-communication-ui/src/calling/java/com/azure/android/communication/ui/calling/configuration/events/CallCompositeEventsHandler.kt
pavelprystinka: It will be up to Contoso to to make it thread safe.
If decide to make it thread safe we can make it later, as it will not introduce a braking change. Now we need to make this change as it will not be possible to change public api for this reason.

azure-communication-ui/azure-communication-ui/src/calling/java/com/azure/android/communication/ui/calling/configuration/events/CallCompositeEventsHandler.kt
pavelprystinka: Thank you, we will consider it in the future. We may also use azure core future implementation for events.

iaulakh:
tested

iaulakh (approved):

ahammer (approved):

azure-communication-ui/azure-communication-ui/src/calling/java/com/azure/android/communication/ui/calling/configuration/events/CallCompositeEventsHandler.kt
waynemo, line 12: `remoteParticipantJoinedHandlers`

azure-communication-ui/azure-communication-ui/src/calling/java/com/azure/android/communication/ui/calling/configuration/events/CallCompositeEventsHandler.kt
```diff
@@ -8,19 +8,23 @@ import com.azure.android.communication.ui.calling.models.CallCompositeErrorEvent
...
+        mutableSetOf<CallCompositeEventHandler<CallCompositeRemoteParticipantJoinedEvent>>()
 
-    fun getOnErrorHandler() = errorHandlers
+    fun getOnErrorHandlers() = errorHandlers.asIterable()
```
waynemo, line 15: Does `.asIterable()` create a new object every time?

waynemo (changes requested):

azure-communication-ui/azure-communication-ui/src/calling/java/com/azure/android/communication/ui/calling/configuration/events/CallCompositeEventsHandler.kt
pavelprystinka: We will fix it.

PR #455, opened 2022-09-28
[Chat][Feature] start stop events notifications

dhiraj:
## Purpose

* This PR adds the ability for developer to start and stop realtime event notifications from Chat SDK

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [x] Feature

## Pull Request Checklist

- [ ] Public API changes
- [ ] UI Changes
- [ ] Tests
  - [ ] Unit Tests Included

## Questions
1. We are silently swallowing any exceptions thrown by the ChatSDK for failing to start event notifications. Should we surface this to developers?
2. Do we need a de-initialize call in ChatSDK so that we can stop event notifications even if developer does not ask for this explicitly?


ShaunaSong (dismissed):

azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/service/sdk/ChatSDK.kt
```diff
@@ -17,4 +18,7 @@ internal interface ChatSDK {
         type: ChatMessageType,
         content: String,
     ): CompletableFuture<SendChatMessageResult>
+
+    fun startEventNotifications(context: Context) {}
```
iaulakh, line 22: We can remove context from here. Wrapper class already have context in constructor.

azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/service/sdk/ChatSDKWrapper.kt
```diff
@@ -73,6 +74,17 @@ internal class ChatSDKWrapper(
         return future
     }
 
+    override fun startEventNotifications(context: Context) {
+        if (startedEventNotifications) return
+        chatAsyncClient.startRealtimeNotifications(context) {}
```
iaulakh, line 79: can we do something like? chatAsyncClient.startRealtimeNotifications(context) {
            if(it == null) {
                // set bool to ture
            } else {
                throw it
            }
        }

azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/service/sdk/ChatSDK.kt
dhiraj: @iaulakh but the context is not being held in a variable so it isn't directly available (outside the constructor), right? I can add an instance variable to hold it and then pass it in, but then we'll probably have to make a de-initialize call to stop holding on to the context otherwise there'll be a memory leak, I think. 🤔

iaulakh: sounds good Dhiraj.
internal class ChatSDKWrapper(
    private val context: Context,
    chatConfig: ChatConfiguration,
)

azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/service/sdk/ChatSDKWrapper.kt
```diff
@@ -73,6 +74,17 @@ internal class ChatSDKWrapper(
...
+        startedEventNotifications = true
```
Mohtasim, line 80: Does startRealTimeNotification have any success callback? I mean can we set true or false within that scope given startRealtimeNotifications(context) {
 // scope
}

azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/service/sdk/ChatSDKWrapper.kt
```diff
@@ -73,6 +74,17 @@ internal class ChatSDKWrapper(
...
+    }
+
+    override fun stopEventNotifications() {
+        chatAsyncClient.stopRealtimeNotifications()
```
Mohtasim, line 84: if there's any scope for stopRealTimeNotification can we set false under that?

azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/service/sdk/ChatSDKWrapper.kt
dhiraj: Using context as instance variable in https://github.com/Azure/communication-ui-library-android/pull/455/commits/093c6e208e9d4ce907bb469abfc93e324bc083e8

azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/service/sdk/ChatSDK.kt
dhiraj: Using context as instance variable in https://github.com/Azure/communication-ui-library-android/commit/093c6e208e9d4ce907bb469abfc93e324bc083e8

azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/service/sdk/ChatSDKWrapper.kt
dhiraj: startEventNotifications() now takes an error handler function as an argument and this will be called with the exception if it happens in https://github.com/Azure/communication-ui-library-android/commit/093c6e208e9d4ce907bb469abfc93e324bc083e8

azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/service/sdk/ChatSDKWrapper.kt
```diff
@@ -72,6 +73,22 @@ internal class ChatSDKWrapper(
         return future
     }
 
+    override fun startEventNotifications(errorHandler: (exception: Throwable) -> Unit) {
+        if (startedEventNotifications) return
+        chatAsyncClient.startRealtimeNotifications(context) {
+            if (it == null) {
+                startedEventNotifications = true
+            } else {
```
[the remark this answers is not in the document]
dhiraj: Good point @iaulakh, I've added it in [2569796](https://github.com/Azure/communication-ui-library-android/pull/455/commits/25697961c7fe89cff8f542ecd1ad6ee09d606944)

Mohtasim (dismissed):

iaulakh (approved):

Mohtasim (approved):

PR #548, opened 2022-10-27
[Feature][Chat] TreeBased implementation performance improvements and test

Mohtasim:
## Purpose

* Search API performance improvement in tree based implementation
* Message repo unit tests 
* Performance measure with unit test

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [x] Feature
- [x] Code style update (formatting, local variables)

## Pull Request Checklist

- [ ] Public API changes
- [ ] UI Changes
- [ ] Tests
  - [x] Unit Tests Included

## How to Test

* Open...
* Click on...

## What to Check
* ...


azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/repository/MessageRepository.kt
```diff
@@ -27,7 +34,34 @@ internal class MessageRepository private constructor(
     override fun getLastMessage(): MessageInfoModel? = writerDelegate.getLastMessage()
 
     companion object {
-        fun createListBackedRepository(): MessageRepository {
+
+        private var instance: MessageRepository? = null
```
ahammer, line 38: no singleton please.

We may need to support multiple threads, and thus multiple message repos at once.

Mohtasim: Even if we have multiple threads for write & read, we would want only one storage for messages. A single source of truth.

Otherwise different instances would make different storages. Then for UI to get the messages, each store needs to be synced. I don't think we should go down this type approach.

Mohtasim: Since composite handles instance creation through service locator and its keeps only one instance right now, we can avoid singleton.

Thanks Adam.

ahammer (approved):

iaulakh (approved):

PR #569, opened 2022-11-03
[Chat] Receive read receipt on presentation layer

lpeixin:
## Purpose
Add changes on presentation layer for receiving read receipt

- [ ] Yes
- [X] No

## Pull Request Type

- [ ] Bugfix
- [X] Feature

## Pull Request Checklist

- [ ] Public API changes
- [ ] UI Changes
- [ ] Tests
  - [ ] Unit Tests Included

## How to Test
join chat thread, send message, from other participants' side to read some message sent by local user, then verify if the correct message being marked as read.

### DIFF
--- a/azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/presentation/ui/chat/components/MessageView.kt
+++ b/azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/presentation/ui/chat/components/MessageView.kt
@@ -154,6 +154,18 @@ private fun BasicChatMessage(viewModel: MessageViewModel) {
                         )
                     }
                 }
+                Column{
+                    if (viewModel.isRead) {
+                        Icon(
+                            painter = painterResource(id = R.drawable.azure_communication_ui_chat_ic_fluent_message_read_10_filled),
+                            contentDescription = "Message Read",
+                            modifier = Modifier.padding(
+                                ChatCompositeTheme.dimensions.messageRead
+                            ),
+                            // tint = ChatCompositeTheme.colors.systemIconColor
+                        )
+                    }
+                }
             }
         }
     }
--- a/azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/presentation/ui/chat/components/MessageView.kt
+++ b/azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/presentation/ui/chat/components/MessageView.kt
@@ -171,7 +171,8 @@ private fun BasicChatMessage(viewModel: MessageViewModel) {
                     contentDescription = "Message Read",
                     modifier = Modifier.padding(
                         ChatCompositeTheme.dimensions.messageRead
-                    )
+                    ),
+                    tint = ChatCompositeTheme.colors.unreadMessageIndicatorBackground
                 )
             }
         }
--- a/azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/presentation/ui/chat/components/MessageView.kt
+++ b/azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/presentation/ui/chat/components/MessageView.kt
@@ -4,6 +4,7 @@
 package com.azure.android.communication.ui.chat.presentation.ui.chat.components
 
 import android.widget.TextView
+import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Box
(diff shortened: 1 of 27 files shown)


azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/presentation/ui/chat/components/MessageView.kt
```diff
@@ -114,47 +116,64 @@ private fun BasicChatMessage(viewModel: MessageViewModel) {
...
                     }
                 }
             }
+            androidx.compose.animation.AnimatedVisibility(visible = viewModel.isRead) {
+                Icon(
+                    painter =
+                    painterResource(
+                        id =
+                        R.drawable.azure_communication_ui_chat_ic_fluent_message_read_10_filled
+                    ),
```
iaulakh, line 170: tent is required to display right color for icon

azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/presentation/ui/viewmodel/MessageViewModel.kt
```diff
@@ -55,7 +58,8 @@ private class InfoModelToViewModelAdapter(
                 thisMessage.createdOn ?: OffsetDateTime.now()
             ),
 
-            isLocalUser = isLocalUser
+            isLocalUser = isLocalUser,
+            isRead = isLocalUser && (thisMessage.id == latestReadMessageTimestamp.toString())
```
iaulakh, line 62: expected  
var messageUpdatedTimestamp: OffsetDateTime? = thisMessage.editedOn ?: thisMessage.createdOn
 isLocalUser && messageUpdatedTimestamp != null && messageUpdatedTimestamp <= latestReadMessageTimestamp

iaulakh (dismissed):
please review comments, currently I do not see icon

I think currently, we keep cache in redux for participant read receipts:
{
[participant1, lastReadMessageTime1]
[participant2, lastReadMessageTime2]
[participant3, lastReadMessageTime3]
[participant4, lastReadMessageTime4]
}

then we get min lastReadMessageTime from above and compare it with every message createdon || editedon. 

if createdon | editedon < min lastReadMessageTime -> then we mark message as read


Peixin Liu Leah Xia
I think we need to remove self and admin user from above map? 
when we add participant to cache, what timestamp we set? (probably last message created or edited on before participant join)?

Leah-Xia-Microsoft:
> please review comments, currently I do not see icon
> 
> I think currently, we keep cache in redux for participant read receipts: { [participant1, lastReadMessageTime1] [participant2, lastReadMessageTime2] [participant3, lastReadMessageTime3] [participant4, lastReadMessageTime4] }
> 
> then we get min lastReadMessageTime from above and compare it with every message createdon || editedon. 
> 
> if createdon | editedon < min lastReadMessageTime -> then we mark message as read
> 
> Peixin Liu Leah Xia I think we need to remove self and admin user from above map?  when we add participant to cache, what timestamp we set? (probably last message created or edited on before participant join)?

 - For the admin user part, we will handle it in this ticket https://skype.visualstudio.com/SPOOL/_workitems/edit/3044060. So for now, unless we hard code the admin user to test, we won't see the read icon

 - I agree that we should exclude self from above map

 - We use the messageId as the timeStamp. MessageId is epoch time itself.

azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/presentation/ui/viewmodel/MessageViewModel.kt
lpx1989: addressed

azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/presentation/ui/chat/components/MessageView.kt
lpx1989: addressed

azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/presentation/ui/viewmodel/MessageViewModel.kt
```diff
@@ -55,7 +59,8 @@ private class InfoModelToViewModelAdapter(
                 thisMessage.createdOn ?: OffsetDateTime.now()
             ),
 
-            isLocalUser = isLocalUser
+            isLocalUser = isLocalUser,
+            isRead = isLocalUser && (currentMessageTime != null && currentMessageTime <= latestReadMessageTimestamp)
```
Mohtasim, line 63: if isLocalUser is true, doesn't it mean that this message is sent by myself? 
In that case, wouldn't isRead be always true? 
So in this case, if isLocalUser is true and currentMessageTime is greater than latestReadMessageTimestamp, isRead should be true too. 
But from the logic written here, it will be false when it should be true.

Mohtasim: my bad. this is receiving condition and it looks fine.

Mohtasim (approved):

ShaunaSong (approved):

PR #620, opened 2022-11-23
[Feature] Code coverage for all projects

grigoryk:
We'd like to generate code coverage reports for all of the subprojects - `calling`, `chat`, `call-with-chat`. Currently, we're only doing this for `calling`.

This adds a new root level task, `combinedCoverageReport`, which knows how to generate a combined coverage report against debug `exec` files in all of the subprojects.

So, to get a full coverage report for debug unit tests:
- Run all unit tests: `./gradlew testDebugUnitTest`
- Generate the combined report: `./gradlew combinedCoverageReport`

Reports (html and xml) will be placed into `$rootProjectDir/build/reports/coverage/test/debug/`.

Reporting task hardcodes looking at coverage data specifically for the `debug` variant. We could fix that in the future if we end up running non-debug unit tests in CI, but for now this seems fine.

I've removed the existing jacoco stuff we had for `calling`. We didn't really need it, it seems - with this PR applied, you can still use standard ways of generating these reports for individual subprojects. E.g., see:

```
./gradlew tasks --all | grep coverage | cut -d " " -f 1
combinedCoverageReport
demo-app:createCallingDebugUnitTestCoverageReport
demo-app:createCallingReleaseUnitTestCoverageReport
demo-app:createCallwithchatDebugUnitTestCoverageReport
demo-app:createCallwithchatReleaseUnitTestCoverageReport
demo-app:createChatDebugUnitTestCoverageReport
demo-app:createChatReleaseUnitTestCoverageReport
call-with-chat:createDebugUnitTestCoverageReport
calling:createDebugUnitTestCoverageReport
chat:createDebugUnitTestCoverageReport
call-with-chat:createReleaseUnitTestCoverageReport
calling:createReleaseUnitTestCoverageReport
chat:createReleaseUnitTestCoverageReport

./gradlew calling:createDebugUnitTestCoverageReport --dry-run
...tasks...
:calling:testDebugUnitTest SKIPPED
:calling:createDebugUnitTestCoverageReport SKIPPED

```


grigoryk:
Note that our CI currently hardcodes running/processing of the old coverage report in calling specifically, so that will need to change as well.

grigoryk:
https://github.com/Azure/communication-ui-library-android/pull/621 transitions our main CI pipeline to YAML, so let's merge that first, and then adjust this PR to modify that pipeline.

grigoryk:
Okay, rebased on top of merged-in #621. Should be good to go once green.

ahammer (dismissed):

grigoryk:
Err, just to be "proper" kept the original jacoco task for now, since the legacy CI pipeline (which hardcodes against it) is still configured until we the yml is available in all the relevant branches. So, both pipelines should be green now.

cc @ahammer sorry for the churn.

Mohtasim (approved):

ShaunaSong (approved):

PR #710, opened 2023-01-30
[Feature][Calling] CSS call history

pavelprystinka:
## Purpose

* CSS call history

- [ ] Yes
- [ ] No

## Pull Request Type

- [ ] Bugfix
- [X] Feature

## Pull Request Checklist

- [ ] Public API changes
- [ ] UI Changes
- [ ] Tests
  - [ ] Unit Tests Included

## How to Test

* Open...
* Click on...

## What to Check
* ...


azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/service/CallHistoryService.kt
```diff
@@ -0,0 +1,54 @@
...
+    fun start(coroutineScope: CoroutineScope)
+}
+
+internal class CallHistoryServiceImpl(
+    private val store: Store<ReduxState>,
+    private val callHistoryRepository: CallHistoryRepository,
+) : CallHistoryService {
+    private val callIdStateFlow = MutableStateFlow<String?>(null)
+
+    override fun start(coroutineScope: CoroutineScope) {
```
grigoryk, line 25: Since this will actually do an `insert`, I would call this something like `recordCallStart` so that it's obvious to the callers what happens. Otherwise this `insert` almost looks like a non-obvious side-effect of "starting a service".

Also there's no corresponding "stop" method for this service, so I'd definitely not using the name "start" for this.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/service/CallHistoryService.kt
```diff
@@ -0,0 +1,54 @@
...
+            callIdStateFlow.collect { callId ->
+                val callStartDateTime = store.getCurrentState().callState.callStartLocalDateTime
+                if (callId != null && callStartDateTime != null) {
+                    callHistoryRepository.insert(callId, callStartDateTime)
+                }
+            }
+        }
+
+        coroutineScope.launch {
+            cleanOldRecords()
```
grigoryk, line 42: I guess it's fine to do the cleanup on every insert, and not just once during `init`.

But, if you're going to do this, I would just hide the cleanup login as part of the insert logic within the repository implementation, and do both INSERT and DELETE wrapped together in a single transaction.

It'll be faster and more robust. Less transaction handling for sqlite to do with a single transaction; as-is, you're implicitly beginning and ending at least 2+N transactions (N=# of records to cleanup).

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,73 @@
...
+                        )
+                    )
+                } while (moveToNext())
+            }
+        }
+        cursor.close()
+        return items
+    }
+
+    override fun remove(id: Int) {
```
grigoryk, line 68: Don't need this

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,73 @@
...
+        val db = DbHelper(context).writableDatabase
+
+        val date = callDateTime.format(DateTimeFormatter.ofPattern(datePattern))
+
+        val values = ContentValues().apply {
+            put(CallHistoryContract.COLUMN_NAME_CALL_ID, callId)
+            put(CallHistoryContract.COLUMN_NAME_CALL_DATE, date)
+        }
+
+        db.insert(CallHistoryContract.TABLE_NAME, null, values)
```
grigoryk, line 34: Do the cleanup here as a single DELETE statement, wrapped in a transaction together with the INSERT.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,73 @@
...
+import android.content.ContentValues
+import android.content.Context
+import com.azure.android.communication.ui.calling.data.model.CallHistoryRecordData
+import org.threeten.bp.LocalDateTime
+import org.threeten.bp.format.DateTimeFormatter
+
+internal interface CallHistoryRepository {
+    fun insert(callId: String, callDateTime: LocalDateTime)
+    fun getAll(): List<CallHistoryRecordData>
+    fun remove(id: Int)
```
grigoryk, line 15: Don't need this, I think.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,73 @@
...
+}
+
+internal class CallHistoryRepositoryImpl(
+    private val context: Context
+) : CallHistoryRepository {
+
+    private val datePattern: String = "yyyy-MM-dd HH:mm:ss"
+
+    override fun insert(callId: String, callDateTime: LocalDateTime) {
+        val db = DbHelper(context).writableDatabase
```
grigoryk, line 25: The way you're using DbHelper isn't optimal here. Since it extends `SQLiteOpenHelper`, it will handle things like caching of an opened db for you.

You're creating this instance, and throwing it away right after. Which means you're throwing away your "opened database cache".

So, every time you're accessing `CallHistoryRepositoryImpl`, you're going through the entire "open database on disk, run the CREATE statements if necessary, check its version pragma, determine if we need to upgrade, etc" logic. A bunch of slow and unnecessary work.

Instead, consider storing `DbHelper(context)` as a variable on this class. This way you're only incurring the costs of opening/setting up the database once.

Also, you can then control when you incur the db open/setup costs - when this class is initialized (e.g., touch `writableDatabase` in an `init {}` block), or on a first interaction, etc.

You'll also need to determine when you want to close this database, as well - if at all. Probably when `CallCompositeActivity` is being torn down?

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        db.insert(CallHistoryContract.TABLE_NAME, null, values)
+        db.close()
+    }
+
+    override fun getAll(): List<CallHistoryRecordData> {
+        val db = DbHelper(context).writableDatabase
+
+        val cursor = db.rawQuery(
```
grigoryk, line 41: Use Kotlin's try-with-resource equivalent, `use`.

`db.rawQuery(...).use { ... }`

This will ensure to call `close` in a finally block of a try/catch. See https://github.com/JetBrains/kotlin/blob/885d99068c35ac7c959a4db29c3e1d543a7459b6/libraries/stdlib/jvm/src/kotlin/io/Closeable.kt#L21

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/DbHelper.kt
```diff
@@ -0,0 +1,42 @@
...
+internal object CallHistoryContract {
+
+    const val ID = BaseColumns._ID
+    const val TABLE_NAME = "call_history"
+
+    const val COLUMN_NAME_CALL_ID = "call_id"
+    const val COLUMN_NAME_CALL_DATE = "call_date"
+
+    const val SQL_CREATE_CALL_HISTORY =
+        "CREATE TABLE $TABLE_NAME (" +
```
grigoryk, line 38: Do a `CREATE TABLE IF NOT EXISTS` just in case. E.g. if somehow `onCreate` is called while we already have a an existing database. There's a small time window when this can happen, I think - `SQLiteOpenHelper` constructor is called and it determines that there's no db on disk yet, and it needs to create one. So, it creates it, calls `onCreated`, but maybe before that can do its thing, OS kills the app.

Then you end up with an empty db file on disk, and next time this runs, we'll permafail - we have the db already, probably with the correct user_version pragma, so no need to call `oncreate`. But, that db is actually empty, without your table in it. And so everything breaks.

Or, maybe, this scenario isn't possible, and this is all moot. But I think it is possible. Either way, appending `IF NOT EXISTS` lets us to not worry about this :-)

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/DbHelper.kt
```diff
@@ -0,0 +1,42 @@
...
+    companion object {
+        // If you change the database schema, you must increment the database version.
+        const val DATABASE_VERSION = 1
+        const val DATABASE_NAME = "com.azure.android.communication.ui.calling.CallHistoryReader.db"
+    }
+}
+
+internal object CallHistoryContract {
+
+    const val ID = BaseColumns._ID
```
grigoryk, line 31: I don't think you need a separate id field, especially once you simplify your cleanup logic to not address individual rows directly, and use a date range instead.

A sqlite db will already have a `rowid` unless you ask it not during table creation (see https://www.sqlite.org/rowidtable.html, esp the apology note at the bottom). So this id, roughly, becomes an alias for that field.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/DbHelper.kt
```diff
@@ -0,0 +1,42 @@
...
+    const val TABLE_NAME = "call_history"
+
+    const val COLUMN_NAME_CALL_ID = "call_id"
+    const val COLUMN_NAME_CALL_DATE = "call_date"
+
+    const val SQL_CREATE_CALL_HISTORY =
+        "CREATE TABLE $TABLE_NAME (" +
+            "${BaseColumns._ID} INTEGER PRIMARY KEY," +
+            "$COLUMN_NAME_CALL_ID TEXT," +
+            "$COLUMN_NAME_CALL_DATE TEXT)"
```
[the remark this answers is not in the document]
grigoryk: Probably worth adding an index on call_date as well, since your cleanup will be doing a range lookup.
`CREATE INDEX IF NOT EXISTS call_dateindex ON tableaname(call_date);`

grigoryk (changes requested):
Did a first pass, mostly just on the db stuff. It mostly looks fine, but see comments around specifics.

In summary:
- better lifecycle of the db connection (don't throw away the opened connection)
- simplify schema (remove `id`)
- simplify cleanup logic, make it part of an insert logic as a single DELETE in a transaction; won't need a `remove` method then
- switch to just using Long for start timestamp. it's more correct, and you won't need to do all that time parsing stuff and won't need a threetenbp library include (or desugaring java8+ features)
- I'm fairly sure you're doing all the db operations on the main dispatcher (due to using lifecycleScope), which is not what you want. Use a dedicated single-threaded executor instead (since sqlite is inherently single-threaded and won't like it if we write simultaneously on different threads).

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/service/CallHistoryService.kt
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+
+    private fun cleanOldRecords() {
+        val thresholdDate = LocalDateTime.now().minusDays(31)
+        callHistoryRepository.getAll().forEach {
```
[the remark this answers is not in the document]
grigoryk: Actually, do it as part of the insert, not a separate method. Don't need we need one.

azure-communication-ui/calling/build.gradle
```diff
@@ -99,12 +99,16 @@ dependencies {
     implementation "com.microsoft.fluentui:fluentui_persona:$microsoft_fluent_ui_version"
     implementation "com.microsoft.fluentui:fluentui_transients:$microsoft_fluent_ui_version"
 
+    api 'com.jakewharton.threetenabp:threetenabp:1.4.4'
```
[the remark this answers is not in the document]
katerinkus: Ah, see comments later - let's just use a `long` instead to store the date. Don't need this nor the desugaring stuff.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,73 @@
...
+                            id = getInt(idColumnIndex),
+                            callId = getString(nameColumnIndex),
+                            date = LocalDateTime.parse(getString(dateColumnIndex), dateFormatter),
+                        )
+                    )
+                } while (moveToNext())
+            }
+        }
+        cursor.close()
+        return items
```
katerinkus, line 65: db.close missing ;) But, let's not close it after each interacting - cache the connection instead.

azure-communication-ui/calling/build.gradle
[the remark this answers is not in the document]
pavelprystinka: threeten is required by Azure guidelines due to old API support requirement: https://azure.github.io/azure-sdk/android_design.html#using-primitive-types

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,73 @@
...
+
+package com.azure.android.communication.ui.calling.data
+
+import android.content.ContentValues
+import android.content.Context
+import com.azure.android.communication.ui.calling.data.model.CallHistoryRecordData
+import org.threeten.bp.LocalDateTime
+import org.threeten.bp.format.DateTimeFormatter
+
+internal interface CallHistoryRepository {
```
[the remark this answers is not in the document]
pavelprystinka: Done

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/DbHelper.kt
pavelprystinka: I think it still makes sense to keep explicit id column due to:

The exception to this rule is when the rowid table declares an INTEGER PRIMARY KEY. In the exception, the INTEGER PRIMARY KEY becomes an alias for the rowid.

Therefore, applications should not normally access the rowid directly, but instead use an INTEGER PRIMARY KEY.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/model/CallHistoryRecordData.kt
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the MIT License.
+
+package com.azure.android.communication.ui.calling.data.model
+
+import org.threeten.bp.LocalDateTime
+
+internal data class CallHistoryRecordData(
+    val id: Int,
+    val callId: String,
+    val date: LocalDateTime,
```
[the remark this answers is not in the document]
pavelprystinka: renamed to callStartedOn (same as the one on the public API

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/DbHelper.kt
pavelprystinka: Add `IF NOT EXISTS`

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
pavelprystinka: Updated

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
pavelprystinka: As per our discussion, decided to keep current logic as it is a small tradeoff in comparison of risk to keep DB connection open, as we do not have a reliable close event.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/CallCompositeExtentions.kt
```diff
@@ -0,0 +1,13 @@
...
+
+package com.azure.android.communication.ui.calling
+
+import android.content.Context
+import com.azure.android.communication.ui.calling.data.CallHistoryRepositoryImpl
+import com.azure.android.communication.ui.calling.presentation.manager.DebugInfoManager
+import com.azure.android.communication.ui.calling.presentation.manager.DebugInfoManagerImpl
+
+internal fun createDebugInfoManager(context: Context): DebugInfoManager {
+    return DebugInfoManagerImpl(CallHistoryRepositoryImpl(context))
```
ahammer, line 12: use context.applicationContext here if possible

pavelprystinka: Good idea, moved context.applicationContext higher in call stack

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
pavelprystinka: removed

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,73 @@
...
+    fun insert(callId: String, callDateTime: LocalDateTime)
```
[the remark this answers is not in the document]
pavelprystinka: I kept function interface as date, but internally stored value changed long. It is up to Repository to decide how to store.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,73 @@
...
+        val cursor = db.rawQuery(
+            "SELECT * FROM ${CallHistoryContract.TABLE_NAME} ORDER BY ${CallHistoryContract.COLUMN_NAME_CALL_DATE} ASC",
+            null
+        )
+
+        val dateFormatter = DateTimeFormatter.ofPattern(datePattern)
+        val items = mutableListOf<CallHistoryRecordData>()
+        with(cursor) {
+            if (moveToFirst()) {
+                val idColumnIndex = getColumnIndex(CallHistoryContract.ID)
```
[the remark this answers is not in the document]
pavelprystinka: Done.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/model/CallHistoryRecordData.kt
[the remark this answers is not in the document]
pavelprystinka: Changed to OffsetDateTime

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/service/CallHistoryService.kt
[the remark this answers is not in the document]
pavelprystinka: Moved clean up logic to insert function of Repository

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/model/CallHistoryRecordData.kt
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.android.communication.ui.calling.data.model
+
+import org.threeten.bp.LocalDateTime
+
+internal data class CallHistoryRecordData(
+    val id: Int,
```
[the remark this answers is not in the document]
pavelprystinka: As per offline discussion decided to keep it in case of future use this table with other tables

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/service/CallHistoryService.kt
```diff
@@ -0,0 +1,54 @@
...
+            store.getStateFlow().collect {
+                callIdStateFlow.value = it.callState.callId
+            }
+        }
+
+        coroutineScope.launch {
+            callIdStateFlow.collect { callId ->
+                val callStartDateTime = store.getCurrentState().callState.callStartLocalDateTime
+                if (callId != null && callStartDateTime != null) {
+                    callHistoryRepository.insert(callId, callStartDateTime)
```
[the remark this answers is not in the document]
pavelprystinka: As this feature is intended to be used for troubleshooting we might not have a better point in time to record start call date. We also do not have any call start date returned from SDK, but even if that would be available that could be not reliable enough in case of failures/issues

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/service/CallHistoryService.kt
pavelprystinka: As per our conversation this method does not do insert right away, it will launch observing on the redux state change.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/service/CallHistoryService.kt
pavelprystinka: Moved.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
pavelprystinka: Removed

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
pavelprystinka: Done

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/DbHelper.kt
pavelprystinka: As per offline meeting decided to keep ID

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/DbHelper.kt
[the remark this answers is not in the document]
pavelprystinka: Added

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/redux/state/CallingState.kt
```diff
@@ -24,6 +26,7 @@ internal data class CallingState(
     val joinCallIsRequested: Boolean = false,
     val isRecording: Boolean = false,
     val isTranscribing: Boolean = false,
+    val callStartLocalDateTime: OffsetDateTime? = null,
```
grigoryk, line 29: rename; also, document that this is fixed for the duration of the call (in a reducer's handling of CallStart event)

grigoryk (commented):
Reviewed in person.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/manager/DebugInfoManager.kt
```diff
@@ -3,35 +3,40 @@
...
+    private suspend fun getCallHistory(): List<CallCompositeCallHistoryRecord> {
+        return callHistoryRepository.getAll()
+            .groupBy {
+                it.callStartedOn
+            }
+            .map { mapped ->
+                buildCallHistoryRecord(mapped.key, mapped.value.map { it.callId })
+            }
+            .sortedBy {
+                it.callStartedOn
```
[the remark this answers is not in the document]
pavelprystinka: I believe yes, as we read denormalized data from DB, groupby or map may change order of items. I will remove ORDER BY from sql query to save resources.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,82 @@
...
+}
+
+internal class CallHistoryRepositoryImpl(
+    private val context: Context
+) : CallHistoryRepository {
+    private val insetLock = Any()
+
+    override suspend fun insert(callId: String, callDateTime: OffsetDateTime) {
+        return withContext(Executors.newSingleThreadExecutor().asCoroutineDispatcher()) {
+            synchronized(insetLock) {
```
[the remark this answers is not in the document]
pavelprystinka: Done.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,82 @@
...
+                DbHelper(context).writableDatabase
```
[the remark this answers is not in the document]
pavelprystinka: Done

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,82 @@
...
+                }
+                return@withContext items
+            }
+        }
+    }
+
+    private fun cleanUpOldRecords(db: SQLiteDatabase) {
+        val threshold = OffsetDateTime.now().minusDays(31).toInstant().epochSecond
+        val sql = "DELETE FROM ${CallHistoryContract.TABLE_NAME} " +
+            "WHERE ${CallHistoryContract.COLUMN_NAME_CALL_DATE} < $threshold"
```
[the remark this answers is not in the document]
pavelprystinka: Done

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeCallHistoryRecord.java
```diff
@@ -0,0 +1,36 @@
...
+package com.azure.android.communication.ui.calling.models;
+
+import java.util.List;
+import org.threeten.bp.OffsetDateTime;
+
+/**
+ * Call history.
+ */
+public class CallCompositeCallHistoryRecord {
+    private final  OffsetDateTime callStartedOn;
```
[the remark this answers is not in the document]
pavelprystinka: Removed

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,82 @@
...
+        return withContext(Executors.newSingleThreadExecutor().asCoroutineDispatcher()) {
+            synchronized(insetLock) {
+                DbHelper(context).writableDatabase
+                    .use { db ->
+                        val values = ContentValues().apply {
+                            put(CallHistoryContract.COLUMN_NAME_CALL_ID, callId)
+                            put(CallHistoryContract.COLUMN_NAME_CALL_DATE, callDateTime.toInstant().epochSecond)
+                        }
+
+                        db.insert(CallHistoryContract.TABLE_NAME, null, values)
```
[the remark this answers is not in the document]
pavelprystinka: Done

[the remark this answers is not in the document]
pavelprystinka: Done

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,82 @@
...
+                        cleanUpOldRecords(db)
+                        db.close()
```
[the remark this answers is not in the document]
pavelprystinka: Removed

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,82 @@
...
+            }
+        }
+    }
+
+    override suspend fun getAll(): List<CallHistoryRecordData> {
+        return withContext(Dispatchers.IO) {
+            DbHelper(context).writableDatabase.use { db ->
+                val items = mutableListOf<CallHistoryRecordData>()
+                db.rawQuery(
+                    "SELECT * FROM ${CallHistoryContract.TABLE_NAME} ORDER BY ${CallHistoryContract.COLUMN_NAME_CALL_DATE} ASC",
```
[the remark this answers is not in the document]
pavelprystinka: Done

azure-communication-ui/calling/src/test/java/com/azure/android/communication/ui/service/CallHistoryServiceTest.kt
```diff
@@ -0,0 +1,67 @@
...
+import org.mockito.kotlin.any
+import org.mockito.kotlin.doAnswer
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.times
+import org.mockito.kotlin.eq
+import org.threeten.bp.OffsetDateTime
+
+@RunWith(MockitoJUnitRunner::class)
+internal class CallHistoryServiceTest : ACSBaseTestCoroutine() {
```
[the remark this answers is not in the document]
pavelprystinka: Tests are added.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/redux/state/CallingState.kt
pavelprystinka: Done

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeDebugInfo.java
```diff
@@ -3,30 +3,24 @@
...
-        return this;
+    CallCompositeDebugInfo(final List<CallCompositeCallHistoryRecord> callHistoryRecord) {
+        this.callHistoryRecord = callHistoryRecord;
     }
 
     /**
-     * Get last call id.
-     * @return {@link String}
+     * Call history.
+     * @return
```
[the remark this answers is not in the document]
pavelprystinka: Done

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/CallCompositeActivity.kt
```diff
@@ -117,7 +117,7 @@ internal class CallCompositeActivity : AppCompatActivity() {
         }
 
         notificationService.start(lifecycleScope)
-        diagnosticsService.start(lifecycleScope)
+        callHistoryService.start(lifecycleScope)
```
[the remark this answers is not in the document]
pavelprystinka: As discussed, in case of multiple CallComposites, we will control a single access with lock over static object

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,100 @@
...
+                            )
+                        } while (it.moveToNext())
+                    }
+                }
+                return@withContext items
+            }
+        }
+    }
+
+    private fun cleanupOldRecords(db: SQLiteDatabase) {
```
grigoryk, line 89: nit: I'd probably mark this as `suspend` to avoid accidentally calling this code outside of the IO dispatcher after some future refactor that shuffles things around.

Or simply inline this code and remove the function entirely.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
```diff
@@ -0,0 +1,100 @@
...
+            }
+        }
+    }
+
+    override suspend fun getAll(): List<CallHistoryRecordData> {
+        return withContext(Dispatchers.IO) {
+            // Using a new db instance instead of caching one as we do not have a
+            // reliable event when to dispose it.
+            DbHelper(context).writableDatabase.use { db ->
+                cleanupOldRecords(db)
```
grigoryk, line 58: This needs to be in a synchronized block (for the same reason you have one for inserts). A delete is also a "write" operation, so same constraints apply.

I would just remove this line though instead of bumping a simple read to be a write transaction. It's not strictly necessary to cleanup here (you're already cleaning up on inserts).

grigoryk (dismissed):
r+ if you address comments on the CallHistoryRepository.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
pavelprystinka: Was not able to make it suspend as it is in use from critical section

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/data/CallHistoryRepository.kt
pavelprystinka: Fixed.

grigoryk (approved):
Looks good to me! I focused on the repository/etc parts in the review. The rest looks okay at a glance.

ahammer (approved):

PR #711, opened 2023-02-08
[Refactor][Chat] Updated Local User Logic

jpeng-ms:
## Purpose
Note: A upcoming PR would merge remote and local participant info models into one which makes this PR much more sense

Previous implementation was to rely on local user removed action to filter out local participant logic in redux layer. The new approach is to move it into data model (service layer) where a flag would be created under "Participant Info Model" (aka. remote participant info model for today) and use it for logic handling.

This refactor would make the logic to be aligned with iOS

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [ ] Feature

## Pull Request Checklist

- [x] Public API changes
- [x] UI Changes
  - [x] Screen captures included
  - [x] Tested for Light/Dark mode
- [ ] Tests
  - [ ] Unit Tests Included

## How to Test
Since currently there's no way to remove a participant. You need to hard code a timer to simulate a SDK event then check if code is working properly

Paste the following code in chat sdk wrapper line 117 and run the demo app.

```
            val local = RemoteParticipantInfoModel(
                CommunicationIdentifier.CommunicationUserIdentifier("<USER_ID>"),
                "test", true)
            val list = ArrayList<RemoteParticipantInfoModel>();
            list.add(local)
            Timer().schedule(timerTask {
                onChatEventReceived(
                    infoModel = ChatEventModel(
                        eventType = ChatEventType.PARTICIPANTS_REMOVED,
                        infoModel = RemoteParticipantsInfoModel(list)
                    ))
            }, 3000)
```

## What to Check
1. make sure system message of "you were removed from the chat" is shown

## Other Information

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| image | image |

### DIFF
--- a/azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/redux/reducer/ParticipantsReducer.kt
+++ b/azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/redux/reducer/ParticipantsReducer.kt
@@ -31,6 +31,7 @@ internal class ParticipantsReducerImpl : ParticipantsReducer {
             is ParticipantAction.ParticipantsRemoved -> {
                 val participantTypingKeys = state.participantTyping.keys
                 val removedParticipants = action.participants.map { it.userIdentifier.id }
+                var hasLocalParticipant = action.participants.any { it.isLocalUser }
                 var participantTyping = state.participantTyping
                 // TODO: improve this logic
                 removedParticipants.forEach { id ->
@@ -40,7 +41,7 @@ internal class ParticipantsReducerImpl : ParticipantsReducer {
 
                 var updatedState = state
 
-                if (action.localParticipantRemoved) {
+                if (hasLocalParticipant) {
                     updatedState = updatedState.copy(
                         localParticipantInfoModel =
                         state.localParticipantInfoModel.copy(isActiveChatThreadParticipant = false)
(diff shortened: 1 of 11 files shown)


azure-communication-ui/chat/src/main/java/com/azure/android/communication/ui/chat/redux/reducer/ParticipantsReducer.kt
```diff
@@ -40,7 +41,7 @@ internal class ParticipantsReducerImpl : ParticipantsReducer {
 
                 var updatedState = state
 
-                if (action.localParticipantRemoved) {
+                if (hasLocalParticipant) {
```
palatter, line 44: Before this if was based on the local participant being removed but now it's based on having the local participant?

jpeng-ms: right, the naming might be confusing, but `localParticipantRemoved` is a boolean variable inside the action. since `hasLocalParticipant` is within the scope of `is ParticipantAction.ParticipantsRemoved { ... } ` so it serves the same meaning syntactically speaking.

jpeng-ms: sorry not sure if I fully understand your question, but please let me know if you need more clarifications

palatter (approved):

JoshuaLai (approved):

iaulakh (approved):

PR #728, opened 2023-03-26
[Calling][Feature] Rooms integration phase 01

Mohtasim:
## Purpose

* Adds capability to join a rooms call as an attendee

- [x] Yes
- [ ] No

## Pull Request Type

- [ ] Bugfix
- [x] Feature

## Pull Request Checklist

- [x] Public API changes
- [ ] UI Changes
- [ ] Tests
  - [ ] Unit Tests Included

## How to Test

* Open demo app
* enter the room number and acs token
* join rooms call

## What to Check
* If user can join an interact in a rooms call


azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/fragment/setup/components/ErrorInfoView.kt
```diff
@@ -76,9 +76,6 @@ internal class ErrorInfoView(private val rootView: View) {
         if (errorMessage.isBlank()) return
         snackBarTextView.text = errorMessage
         snackBar.run {
-            if (isShown) {
```
Mohtasim, line 79: Why was this removed?

pavelprystinka: If snack bar is already shown, then it does close it, but a new one is not opened. 
From user perspective we do not need to close and open, the only what needs to change is a message.

pavelprystinka (approved):

ShaunaSong (approved):

PR #747, opened 2023-04-12
[Calling][Feature] API to hangup call and call state change notification

iaulakh:
## Purpose

* api to hangup call

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [x] Feature

## Pull Request Checklist

- [x] Public API changes
- [ ] UI Changes
- [x] Tests
  - [x] Memory leak analysis performed
  - [x] Tested on API 21, 26 and latest 
  - [x] Unit Tests Included


azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeExitEvent.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+public final class CallCompositeExitEvent {
+    private final CallCompositeErrorEvent errorEvent;
+
+    /**
+     * Create {@link CallCompositeExitEvent} with error event.
+     *
+     * @param errorEvent error event.
+     */
+    public CallCompositeExitEvent(final CallCompositeErrorEvent errorEvent) {
```
pavelprystinka, line 17: It is weird that we have CallCompositeErrorEvent inside of another event. Would it be better to have an object or enum like ExitCause - that would be an user initiated, API (exit() function call) and error. 
If it is an error cause, then we populate Throwable and error code.

Should we also have a no parameters constructor?

pavelprystinka (changes requested):

pavelprystinka (dismissed):

PR #840, opened 2023-10-04
[Calling][Feature] Teams meeting lobby admit

iaulakh:
## Purpose

* Add UI Support to admit/decline lobby user in teams meeting.

image
image

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [x] Feature

## Pull Request Checklist

- [ ] Public API changes
- [x] UI Changes
  - [x] Screen captures included
  - [x] Tested for Light/Dark mode
  - [x] Tested for screen rotation
  - [x] Tested on Tablet and small screen device (5")
  - [x] Include localization changes
- [x] Tests
  - [x] Memory leak analysis performed
  - [x] Tested on API 21, 26 and latest 
  - [x] Tested for background mode
  - [x] Unit Tests Included

## How to Test

* Join teams meeting and add few lobby users


azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/redux/state/LocalUserState.kt
```diff
@@ -78,4 +79,5 @@ internal data class LocalUserState(
         false,
         false
     ),
+    val localParticipantRole: CallCompositeInternalParticipantRole?
```
pavelprystinka, line 82: Should we use CallCompositeInternalParticipantRole.UNINITIALIZED as a default value instead of null?

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/service/sdk/CallingSDKWrapper.kt
```diff
@@ -56,6 +62,7 @@ internal class CallingSDKWrapper(
 
     private var videoDevicesUpdatedListener: VideoDevicesUpdatedListener? = null
     private var camerasCountStateFlow = MutableStateFlow(0)
+    private var lobby: Lobby? = null
```
pavelprystinka, line 65: Do we need to keep a lobby object here? It can always be accessed from nullableCall?.lobby
This way we will not have a chance to leak this object, as now it is not cleaned.

iaulakh: resolved

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/redux/state/LocalUserState.kt
iaulakh: I think null will be good Pavel, native sdk set to UNINITIALIZED

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/fragment/calling/participantlist/ParticipantListView.kt
```diff
@@ -138,24 +141,35 @@ internal class ParticipantListView(
     }
 
     private fun updateRemoteParticipantListContent(listSize: Int) {
+
+        // title for in call participants
+        var titles = 1
+
+        // title for in lobby participants
+        if (viewModel.getRemoteParticipantListCellStateFlow().value?.any { it.status == ParticipantStatus.IN_LOBBY } == true) {
```
pavelprystinka, line 149: We need also filter out DISCONNECTED users. 
If remote participant is declined to join, then SDK fires event that status is changed to disconnected before removing user from the list.
Same filter needs to be applied to grid as well

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/redux/middleware/handler/CallingMiddlewareActionHandler.kt
```diff
@@ -320,15 +354,15 @@ internal class CallingMiddlewareActionHandlerImpl(
 
     private fun subscribeCamerasCountUpdate(store: Store<ReduxState>) {
         coroutineScope.launch {
-            callingService.getCamerasCountStateFlow().collect {
+            callingService.getCamerasCountStateFlow()?.collect {
```
pavelprystinka, line 357: Why do we need "?" null check here (and five more places in the file)? Returned type by callingService.getCamerasCountStateFlow() is not nullable

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/service/sdk/CallingSDKWrapper.kt
```diff
@@ -142,6 +149,63 @@ internal class CallingSDKWrapper(
         return endCallCompletableFuture!!
     }
 
+    override fun admitAll(): CompletableFuture<CallCompositeLobbyErrorCode?> {
+        val future = CompletableFuture<CallCompositeLobbyErrorCode?>()
+        val options = AdmitLobbyParticipantOptions()
+        nullableCall?.lobby?.admitAll(options)?.whenComplete { _, error ->
```
pavelprystinka, line 155: That should not happen, but what If nullableCall?.lobby? will actually be null, then we will never complete future. 
Should it be completed with error?

pavelprystinka (approved):

ShaunaSong (approved):

PR #894, opened 2024-01-05
Feature Flag Support

ahammer:
## Purpose

* Compile time flag framework for Android

First Step Only.
Support for Beta SDK Version, and Support for Folder based Feature Flags/Implementations.

Can be used for runtime checking, or for compile time splitting (based on API).

Made "CI Compatible", but will need to update CI to improve further (i.e. adding flavors instead of a config).

- [ ] Yes
- [ ] No

## Pull Request Type

- [ ] Bugfix
- [ ] Feature

## Pull Request Checklist

- [ ] Public API changes
- [ ] UI Changes
- [ ] Tests
  - [ ] Unit Tests Included

## How to Test

* Open...
* Click on...

## What to Check
* ...

### DIFF
--- a/azure-communication-ui/build.gradle
+++ b/azure-communication-ui/build.gradle
@@ -32,11 +32,39 @@ buildscript {
         if (project.rootProject.file('private_drop.properties').canRead()) {
             Properties privateDropProps = new Properties()
             privateDropProps.load(project.rootProject.file('private_drop.properties').newDataInputStream())
-            azure_calling_sdk_version = privateDropProps.getProperty('PRIVATE_DROP_VERSION', '""')
+            azure_calling_sdk_version_ga = privateDropProps.getProperty('PRIVATE_DROP_VERSION', '""')
+            azure_calling_sdk_version_beta = azure_calling_sdk_version_ga
         }
         else {
-            azure_calling_sdk_version = '2.6.0-beta.6'
+            // To use Beta Calling SDK and Features, set the following in local.properties:
+            //
+            // USE_CALLING_SDK_BETA=true
+            //
+            // and then run ./gradlew clean
+            // This will use the beta version of the SDK
+            // and enable the beta features
+
+            // In this branch, it was already Beta version, so it'll be in the GA field here
+            // This should be set to a GA release before merge back
+            azure_calling_sdk_version_ga = '2.6.0-beta.6'
+            azure_calling_sdk_version_beta = '2.6.0-beta.6'
         }
+
+        // Initialize the SDK Version based on the calling SDK Beta flag
+        Properties localProps = new Properties()
+        if (project.rootProject.file('local.properties').canRead()) {
+            localProps.load(project.rootProject.file('local.properties').newDataInputStream())
+        }
+
+        // Select GA first, but if that isn't right then BETA
+        ext.azure_calling_sdk_version = ext.azure_calling_sdk_version_ga
+        isUsingBetaSdk=false
+        if (localProps.getProperty('USE_CALLING_SDK_BETA', 'false').toBoolean()) {
+            ext.azure_calling_sdk_version = ext.azure_calling_sdk_version_beta
+            isUsingBetaSdk=true
+        }
+
+
 
         // while modifying this version, make sure to modify in ChatContainer
         azure_chat_sdk_version = '2.0.0'
--- a/azure-communication-ui/calling/README.md
+++ b/azure-communication-ui/calling/README.md
@@ -46,7 +46,14 @@ repositories {
 ```
 Sync project with gradle files. (Android Studio -> File -> Sync Project With Gradle Files)
 
+### Beta Calling SDK and Features
 
+To utilize Beta Calling SDKs and enable features that are in preview, add the following lines to your project level `local.properties` file.
+```properties
+USE_BETA_CALLING_SDK=true
+```
+
+```groovy
 ### Quick Sample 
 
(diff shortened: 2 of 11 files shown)


iaulakh (dismissed):

Mohtasim (approved):

ShaunaSong (approved):

azure-communication-ui/build.gradle
```diff
@@ -32,12 +32,40 @@ buildscript {
...
+            //
+            // USE_CALLING_SDK_BETA=true
+            //
+            // and then run ./gradlew clean
+            // This will use the beta version of the SDK
+            // and enable the beta features
+
+            // In this branch, it was already Beta version, so it'll be in the GA field here
+            // This should be set to a GA release before merge back
+            azure_calling_sdk_version_ga = '2.6.0-beta.6'
```
pavelprystinka, line 49: If we read from local.properties anyway to find out which SDK version to use, why not just keep a single azure_calling_sdk_version variable and read from local.properties an override the azure_calling_sdk_version defined in build.gradle.
So if we add azure_calling_sdk_version to the local.propertyies, then use that value.

pavelprystinka, line 49: If we read from local.properties anyway to find out which SDK version to use, why not just keep a single azure_calling_sdk_version variable and read from local.properties an override the azure_calling_sdk_version defined in build.gradle.
So if we add azure_calling_sdk_version to the local.propertyies, then use that value.

azure-communication-ui/calling/README.md
```diff
@@ -46,7 +46,14 @@ repositories {
 ```
 Sync project with gradle files. (Android Studio -> File -> Sync Project With Gradle Files)
 
+### Beta Calling SDK and Features
```
pavelprystinka, line 49: I do not think we need to share this with Contoso.
They just can go and put any version they would like to test with in the build.gradle.
USE_BETA_CALLING_SDK does not necessary use actual lates released SDK beta.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/features/ACSFeature.kt
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @author yassirb@microsoft.com
+ */
+internal interface ACSFeature {
+    /**
+     * Return feature availability
+     *
+     * @return true if available on the current release otherwise false
+     */
+    fun IsAvailable(): Boolean
```
pavelprystinka, line 17: should it be `isAvailable` get only property?

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/features/ACSFeatureFactory.kt
```diff
@@ -0,0 +1,81 @@
...
+     * @param _instance feature instance
+     * @param <I> type of the feature interface
+     * @param <O> type of the feature instance
+     </O></I> */
+    private inline fun <reified I : ACSFeature> RegisterACSFeature(_instance: I) {
+        featureList!![I::class.java] = _instance
+    }
+
+    companion object {
+        private var featureList: MutableMap<Class<*>, ACSFeature>? = null
```
pavelprystinka, line 74: Why not to make it non-nullable and give is a value` = HashMap()`. Then we will not need to use `!!`

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/features/interfaces/ISupportFilesFeature.kt
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT License.
+
+package com.azure.android.communication.ui.calling.features.interfaces
+
+import android.content.Context
+import com.azure.android.communication.calling.CallClient
+import com.azure.android.communication.ui.calling.features.ACSFeature
+import java.io.File
+
+abstract class ISupportFilesFeature : ACSFeature {
```
pavelprystinka, line 11: It reminds me a C# code style with prefixing `I` for interfaces.
Should we rather use `SupportFilesFeature` for interface and `SupportFilesFeatureimpl` for class.
https://kotlinlang.org/docs/coding-conventions.html#function-names

pavelprystinka (changes requested):

azure-communication-ui/build.gradle
ahammer: it's meant to be pinned to latest public release (with manual updates to the file as we need them). But down the road I can add an over-ride version in local.properties.

azure-communication-ui/calling/README.md
ahammer: Removed from Readme

azure-communication-ui/build.gradle
ahammer: meant to be pinned to latest public release, aligning with Calling SDK
i.e. they always have a latest GA and latest Beta. This should generally be these values here.

We can add a version over-ride to local.properties if people can use. but not in this PR.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/features/ACSFeature.kt
ahammer: will update, I pulled this code direct from calling native

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/features/ACSFeatureFactory.kt
ahammer: will update. Code was grabbed straight from native and went through the Kotlin convertor. I cleaned it up a bit and removed !!'s etc. Will be pushed soon.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/features/interfaces/ISupportFilesFeature.kt
ahammer: Will update.

pavelprystinka (dismissed):

pavelprystinka (approved):

iaulakh (approved):

ShaunaSong (approved):

PR #951, opened 2024-02-28
[Calling][Accessibility][Android][Group/Teams meeting-> Call Setup Page]: Keyboard focus is not visible on the 'Back' button of the 'Setup' call screen.

iaulakh:
## Purpose

* Fix icon for back button, and set as focusable

Before - on focus
image

After code changes - on focus
image

- [ ] Yes
- [x] No

## Pull Request Type

- [x] Bugfix
- [ ] Feature


azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/CallCompositeActivity.kt
```diff
@@ -283,6 +289,28 @@ internal class CallCompositeActivity : AppCompatActivity() {
         }
     }
 
+    private fun actionBarAccessibilitySetup() {
+        try {
+            findViewById<View>(R.id.action_bar)?.let { actionBarView ->
```
ahammer, line 294: Why <View> shouldn't be <ActionBarView>

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/CallCompositeActivity.kt
```diff
@@ -283,6 +289,28 @@ internal class CallCompositeActivity : AppCompatActivity() {
         }
     }
 
+    private fun actionBarAccessibilitySetup() {
+        try {
```
ahammer, line 293: Is try/catch on the entire block necessary?

I get there is a ton of places this can crash
(unchecked casts, .first() on finding the buttonView, etc).

It seems like it relies heavily on magic implementation details (i.e. is the AppCompatImageButton used, is it the first in the list of children, etc).

ahammer: Would moving to Toolbar help with a11y of this issue, I remember Mohtassim saying that it would for some of the other a11y issues like title/subtitle.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/CallCompositeActivity.kt
iaulakh: Its type is ToolBar, do we need to specifically indicate type when we just need to get only child?

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/CallCompositeActivity.kt
iaulakh: @ahammer this code is hack to support accessibility focus. Currently we have only one button and if we need to add more, we need to setup custom action bar. 

Started a discussion to use custom action bar to support accessibility. -> Try catch is added to all as this piece of code is good if works, in any case we are not logging errors thus are only for debug purposes.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/CallCompositeActivity.kt
ahammer: val viewGroup = actionBarView as ViewGroup
If you want ViewGroup, then pass ViewGroup into findViewById

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/CallCompositeActivity.kt
```diff
@@ -283,6 +289,28 @@ internal class CallCompositeActivity : AppCompatActivity() {
...
+            findViewById<View>(R.id.action_bar)?.let { actionBarView ->
+                val viewGroup = actionBarView as ViewGroup
+                val buttonView = viewGroup.children.filter { it is AppCompatImageButton }.first()
+                buttonView.setOnKeyListener { _, keyCode, event ->
```
ahammer, line 297: this buttonView is a guess based on implementation of AppBarView, it could be the wrong button if they change implementation. Which would mean it could apply a keyListener to the wrong button.

PR #963, opened 2024-03-14
[Calling] fix API feedback

iaulakh:
## Purpose

* use disable instead of hide
* use TelecomManager


azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositePushNotification.java
```diff
@@ -20,12 +20,12 @@ public class CallCompositePushNotificationInfo {
     private final String eventTypeResult;
 
     /**
-     * Create {@link CallCompositePushNotificationInfo}.
-     * @param notificationInfoMap Notification info map.
+     * Create {@link CallCompositePushNotification}.
+     * @param notification Notification info map.
```
ahammer, line 24: double notification

map format not clear, i.e. what is expectation of key values? any early validation?

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositePushNotification.java
```diff
@@ -20,12 +20,12 @@ public class CallCompositePushNotificationInfo {
...
      */
-    public CallCompositePushNotificationInfo(final Map<String, String> notificationInfoMap) {
+    public CallCompositePushNotification(final Map<String, String> notification) {
         final com.azure.android.communication.calling.PushNotificationInfo pushNotificationInfo =
-                com.azure.android.communication.calling.PushNotificationInfo.fromMap(notificationInfoMap);
+                com.azure.android.communication.calling.PushNotificationInfo.fromMap(notification);
```
ahammer, line 28: no import here? fully resolved?

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/CallComposite.java
```diff
@@ -181,10 +181,10 @@ public void launch(final Context context,
     /**
      * Handle incoming call push notification.
      * Build composite using {@link CallCompositeBuilder#build(Context, CommunicationTokenCredential)}.
-     * @param info     The {@link CallCompositePushNotificationInfo}.
+     * @param info     The {@link CallCompositePushNotification}.
      * @return {@link CompletableFuture} of {@link Void}.
      */
-    public CompletableFuture<Void> handlePushNotification(final CallCompositePushNotificationInfo info) {
+    public CompletableFuture<Void> handlePushNotification(final CallCompositePushNotification info) {
```
pavelprystinka, line 187: param name should not be info any more. `pushNotification`?

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/configuration/CallCompositeConfiguration.kt
```diff
@@ -10,14 +10,14 @@ import com.azure.android.communication.ui.calling.models.CallCompositeCallScreen
...
-import com.azure.android.communication.ui.calling.models.CallCompositeTelecomOptions
+import com.azure.android.communication.ui.calling.models.CallCompositeTelecomManagerOptions
 
 internal class CallCompositeConfiguration {
     var callScreenOptions: CallCompositeCallScreenOptions? = null
     var context: Context? = null
     var displayName: String? = null
     var credential: CommunicationTokenCredential? = null
-    var telecomOptions: CallCompositeTelecomOptions? = null
+    var telecomOptions: CallCompositeTelecomManagerOptions? = null
```
pavelprystinka, line 20: How about
telecomOptions -> telecomManagerOptions

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositePushNotification.java
iaulakh: Hi Adam, this is for MAUI. We do not want to have native SDK import in public API. For native SDK because of retrofuture we can not generate bindings, thus having native SDK public API will cause UI bindings to fail. 
+ When testing on cross platform, if we see issue we may need to use reflection's.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositePushNotification.java
iaulakh: Hi Adam, the first notification is param name. second is intro text. 

When we receive push from firebase, it is map, we pass it to native SDK. It is is valid it returns valid 'type'. While working on implementation, I will check how native SDK respond to invalid or empty map. Most probably it should return empty or null values.

pavelprystinka (approved):

ShaunaSong (dismissed):

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositePushNotification.java
ahammer: It should still document the expected map format, or link to external documentation that explains it.

end of the day it's still a map and a user should know what is in that map. I.e. if they are writing tests around the feature or something they may choose to mock this.

Regarding param `notification` you can elaborate on it there, i.e.
Firebase cloud messaging notification map (for more information see: [LINK])

It is viable that a customer may not use firebase and another notification mechanism, but still may want to use this code, so providing more context/information is valuable to an end user.

ahammer (dismissed):

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeTelecomManagerOptions.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public final class CallCompositeTelecomManagerOptions {
+    private final CallCompositeTelecomManagerIntegration telecomIntegration;
+
+    /**
+     * Create {@link CallCompositeTelecomManagerOptions}.
+     *
+     * @param telecomIntegration telecom integration.
+     */
+    public CallCompositeTelecomManagerOptions(final CallCompositeTelecomManagerIntegration telecomIntegration) {
```
Mohtasim, line 17: Can this be null? And what would be the fallback strategy there?

Mohtasim (approved):

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeTelecomManagerOptions.java
iaulakh: This is Enum.

pavelprystinka (approved):

ahammer (approved):

PR #992, opened 2024-05-07
[Calling][Feature] Telecom manager integration from native calling SDK

iaulakh:
## Purpose

* API to support integration from native SDK
* Event for audio selection changed to support Contoso for application integration telecom manager
* Updated API to 26

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [x] Feature

## Pull Request Checklist

- [x] Public API changes
- [ ] UI Changes
- [x] Tests
  - [x] Unit Tests Included


azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeTelecomManagerIntegration.java
```diff
@@ -0,0 +1,41 @@
...
+package com.azure.android.communication.ui.calling.models;
+
+import com.azure.android.core.util.ExpandableStringEnum;
+
+import java.util.Collection;
+
+/**
+ * CallCompositeTelecomManagerIntegration {@link CallCompositeTelecomManagerIntegration}.
+ */
+public final class CallCompositeTelecomManagerIntegration
```
pavelprystinka, line 13: Should we consider calling a mode: CallCompositeTelecomManagerIntegrationMode ?

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeTelecomManagerOptions.java
```diff
@@ -0,0 +1,46 @@
...
+     *
+     * @param telecomIntegration The telecom integration.
+     * @param phoneAccountId A string identifier that is unique across PhoneAccountHandles with the
+     *                       same component name. Apps registering PhoneAccountHandles should ensure
+     *                       that the ID provided does not expose personally identifying information.
+     *                       A ConnectionService should use an opaque token as the PhoneAccountHandle identifier.
+     *                       Note: Each String field is limited to 256 characters.
+     */
+    public CallCompositeTelecomManagerOptions(final CallCompositeTelecomManagerIntegration telecomIntegration,
+                                              final String phoneAccountId) {
```
pavelprystinka, line 24: `phoneAccountId` should not be a required field. If Contoso using `APPLICATION_IMPLEMENTED_TELECOM_MANAGER` then they do not need to provide `phoneAccountId`

pavelprystinka (approved):

pavelprystinka (approved):

azure-communication-ui/build.gradle
```diff
@@ -46,10 +46,10 @@ buildscript {
             // and enable the beta features
 
             // Latest GA version of the SDK
-            azure_calling_sdk_version_ga = '2.7.0'
+            azure_calling_sdk_version_ga = '2.9.0-beta.1'
```
ahammer, line 49: I'd encourage setting these correctly and not putting Beta here.

If beta is required for features, ideally use flagging if at all possible so that we can continue to ship on latest GA at any moment.

I do have some compassion because these features are targeted for release, and started prior to flags being widely, but at some point in the future, I'll be much more strict about Beta api's being used in the GA line.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/CallCompositeBuilder.java
```diff
@@ -94,6 +97,18 @@ public CallCompositeBuilder callScreenOptions(final CallCompositeCallScreenOptio
         return this;
     }
 
+    /***
```
ahammer, line 100: nit: /** not /***

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeAudioSelectionChangedEvent.java
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.android.communication.ui.calling.models;
+
+/**
+ * Event with audio selection changed.
+ */
+public final class CallCompositeAudioSelectionChangedEvent {
```
ahammer, line 9: Audio is a bit vague, given there is both Audio Inputs (Microphones) and Audio Outputs (Speakers). But I do acknowledge these are often bundled in the context of mobile (I.e. we aren't saying USB Mic + BT Speakers, not headset on mobile).

While I know we bundle, it might make sense to decouple Input/Output here in the API, as in Teams native there is robust input/output selections, and it might be something we want to replicate in the future.

I.e. 
CallCompositeAudioSelectionChangedEvent {
  InputDevice/OutputDevice
}

Even if that means that they are == for now in the native SDK's, it's clear where audio is coming and going to in both directions, and leaves headroom for more advanced audio routing in the future.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeAudioSelectionMode.java
```diff
@@ -0,0 +1,46 @@
...
+package com.azure.android.communication.ui.calling.models;
+
+import com.azure.android.core.util.ExpandableStringEnum;
+
+import java.util.Collection;
+
+/**
+ * Defines values for CallCompositeAudioSelectionMode.
+ */
+public final class CallCompositeAudioSelectionMode extends ExpandableStringEnum<CallCompositeAudioSelectionMode> {
```
ahammer, line 13: Would recommend both AudioSink (speakers) and AudioSource (microphones) as separate enums, even if they share the same values in some places.

I.e. 
Phone Mic/Phone Loud Speaker
Phone Mic/Phone Headset Speaker
Bluetooth Mic/Bluetooth Headset Speaker

and potentially in the future support for more mode combinations, i.e. like teams.

USB Mic/BT A2DP Speaker
USB MIC/Phone Speaker

etc.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeTelecomManagerIntegrationMode.java
```diff
@@ -0,0 +1,41 @@
...
+package com.azure.android.communication.ui.calling.models;
+
+import com.azure.android.core.util.ExpandableStringEnum;
+
+import java.util.Collection;
+
+/**
+ * CallCompositeTelecomManagerIntegrationMode {@link CallCompositeTelecomManagerIntegrationMode}.
+ */
+public final class CallCompositeTelecomManagerIntegrationMode
```
ahammer, line 13: Expectations for more values here?

I've definitely argued always enum, but ARB may ask, why not a boolean if only two values. 

Also, could identify in this value the default mode in the documentation.

azure-communication-ui/build.gradle
iaulakh: Yes Adam, currently merging to my feature branch only. Expected GA is around next week for TelecomManager SDK. Before merging to develop if not in GA, will add feature flag.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeAudioSelectionChangedEvent.java
iaulakh: I think we had discussion before Adam, SDK as well as WEB use audio as default for output. Whenever input is required mic is indicated. Will book a meeting to discuss with Pavel and finalize. 

Native SDK just used
```
setTelecomManagerAudioRoute
```

Android OS
```
CallAudioState
```

If required we can add output may be? but seems like Audio is by default indicated as output.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeTelecomManagerIntegrationMode.java
iaulakh: Adam, I think we decided always to use Enum not boolean for extensibility. + It have two modes so Enum is perfect.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeAudioSelectionChangedEvent.java
pavelprystinka: I see a point in making it more clear with input/output, but on the other hand we can just follow existing name convention Android is using.

ahammer (approved):

PR #1000, opened 2024-05-16
[Calling][Task] fix ARB feedback for 1.8.0

iaulakh:


ShaunaSong (approved):

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeCallStateChangedEvent.java
```diff
@@ -16,11 +16,15 @@ public final class CallCompositeCallStateChangedEvent {
      * Create {@link CallCompositeCallStateChangedEvent} with call state.
      *
      * @param code call state {@link CallCompositeCallStateCode}.
-     * @param callEndReasonCode the call end reason code.
-     * @param callEndReasonSubCode the call end reason sub code.
-     * @param callId the call id.
      */
-    public CallCompositeCallStateChangedEvent(final CallCompositeCallStateCode code,
+    public CallCompositeCallStateChangedEvent(final CallCompositeCallStateCode code) {
```
pavelprystinka, line 20: I believe it was discussed that we do not need public constructor?

iaulakh: We can not remove existing one :)

PR #1004, opened 2024-05-22
[Calling][Feature] 1 to 1 calling with telecom manager support

iaulakh:
## Purpose

* 1 to 1 calling
* Telecom manager integration

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [x] Feature

## Pull Request Checklist

- [x] Public API changes
- [x] UI Changes
- [x] Tests
  - [ ] Unit Tests Included


Mohtasim (dismissed):

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/ParticipantInfoModel.kt
```diff
@@ -15,7 +15,7 @@ internal enum class ParticipantStatus {
 }
 
 internal data class ParticipantInfoModel(
-    val displayName: String,
+    var displayName: String,
```
ahammer, line 18: ParticipantInfoModel is in redux state right?

Everything should technically be val (Redux state should be immutable)

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/fragment/calling/CallingViewModel.kt
```diff
@@ -30,6 +31,7 @@ internal class CallingViewModel(
     private val callScreenOptions: CallCompositeCallScreenOptions? = null,
     val multitaskingEnabled: Boolean,
     val avMode: CallCompositeAudioVideoMode,
+    private val callType: CallType? = null,
```
ahammer, line 34: Possible for a non-null default here? I.e. CallType.Undefined.

iaulakh: Hi Adam, We do not have call type undefined. This should never be null, As on launch call type is set. However as config object is somehow nullable(should not be), thus it forces to set type null. I think I will create a KTLO to remove null from call config as it can never be null.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/ParticipantInfoModel.kt
iaulakh: Adam, info models are created in service layer and cached. When display name changed(can change while in call), info model is updated thus few are var.

Redux is also keeping list but we have copy in middleware I think. If required we can separate object for redux and service layer.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeIncomingCallCancelledEvent.java
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.android.communication.ui.calling.models;
+
+/**
+ * Incoming call cancelled event.
+ */
+public class CallCompositeIncomingCallCancelledEvent {
```
pavelprystinka, line 9: Should be final

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositePushNotification.java
```diff
@@ -0,0 +1,110 @@
...
+
+import com.azure.android.communication.common.CommunicationIdentifier;
+import com.azure.android.communication.ui.calling.CallCompositeException;
+
+import java.util.Map;
+
+/**
+ * CallCompositePushNotification for forwarding calling push notifications to UI Library.
+ */
+public class CallCompositePushNotification {
```
pavelprystinka, line 14: Should be final

pavelprystinka (changes requested):

pavelprystinka (approved):

pavelprystinka (approved):

Mohtasim (approved):

ahammer (approved):

PR #1064, opened 2024-08-16
[Calling][Feature] call header customization for title and timer 

iaulakh:
## Purpose

* Call header title customization
* Timer for call duration
* Participant removed event

- [ ] Yes
- [ ] No

## Pull Request Type

- [ ] Bugfix
- [ ] Feature

## Pull Request Checklist

- [ ] Public API changes
- [ ] UI Changes
- [ ] Tests
  - [ ] Unit Tests Included

## How to Test

* Open...
* Click on...

## What to Check
* ...


azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/configuration/CallCompositeConfiguration.kt
```diff
@@ -41,4 +43,11 @@ internal class CallCompositeConfiguration {
     var capabilitiesChangedNotificationMode: CallCompositeCapabilitiesChangedNotificationMode? = null
     var setupScreenOptions: CallCompositeSetupScreenOptions? = null
     var identifier: CommunicationIdentifier? = null
+
+    fun setTimerConfig(timerConfig: CallDurationManager) {
```
pavelprystinka, line 47: I do not think Configuration should have any business logic in it. It only should have a configuration objects.
Also `setTimerConfig` does not reflect what it does.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/CallComposite.java
```diff
@@ -804,14 +838,18 @@ private void launchComposite(final Context context,
                 participants,
                 incomingCallId));
 
+        final CallDurationManager durationManager = new CallDurationManager();
+        configuration.setTimerConfig(durationManager);
```
pavelprystinka, line 842: Why do we start timer when call composite is launched, instead of starting it on call connecting status change?

pavelprystinka, line 842: We should not set manager to the config object

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeRemoteParticipantLeftEventExtension.kt
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
pavelprystinka, line 1: Please re-use existing `CallCompositeExtentions`. I think we should not create so many small files to put extentions in.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/PackagePrivate.kt
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
pavelprystinka, line 1: Same comment here about `CallCompositeExtentions`

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/CallCompositeActivity.kt
```diff
@@ -214,6 +215,7 @@ internal open class CallCompositeActivity : AppCompatActivity() {
 
             if (isFinishing && store.getCurrentState().navigationState.navigationState == NavigationStatus.EXIT) {
                 store.dispatch(CallingAction.CallEndRequested())
+                callDurationManager.onReset()
```
pavelprystinka, line 218: We can not bind any call logic to the activity lifecycle.
It needs to be done in the Middleware based on the call status change or on `CallEndRequested`.

pavelprystinka (changes requested):

ShaunaSong (approved):

PR #1072, opened 2024-08-28
[Calling][Feature] Call screen information header title, subtitle

iaulakh:
## Purpose

* Title update for call information header
* Subtitle update for call information header

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [x] Feature

## Pull Request Checklist

- [x] Public API changes
- [x] UI Changes
- [x] Tests
  - [ ] Unit Tests Included


azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/redux/action/CallScreenInformationHeaderAction.kt
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/* <CUSTOM_CALL_HEADER> */
+package com.azure.android.communication.ui.calling.redux.action
+
+internal sealed class CallScreenInformationHeaderAction : Action {
```
pavelprystinka, line 7: How about a bit shorter name: `CallScreenInfoHeaderAction`

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/redux/reducer/CallScreenInformationHeaderReducer.kt
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+/* <CUSTOM_CALL_HEADER> */
+package com.azure.android.communication.ui.calling.redux.reducer
+
+import com.azure.android.communication.ui.calling.redux.action.Action
+import com.azure.android.communication.ui.calling.redux.action.CallScreenInformationHeaderAction
+import com.azure.android.communication.ui.calling.redux.state.CallScreenInformationHeaderState
+
+internal interface CallScreenInformationHeaderReducer : Reducer<CallScreenInformationHeaderState>
```
pavelprystinka, line 10: How about `CallScreenInfoHeaderReducer`

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/models/CallCompositeCallScreenHeaderOptions.java
```diff
@@ -3,12 +3,15 @@
...
 package com.azure.android.communication.ui.calling.models;
 
+import com.azure.android.communication.ui.calling.presentation.manager.CallScreenInfoHeaderManager;
+
 /**
  * Options for the CallCompositeCallScreenHeaderOptions.
  */
 public final class CallCompositeCallScreenHeaderOptions {
-    private CallCompositeCallDurationTimer timer;
+    CallScreenInfoHeaderManager callScreenInfoHeaderManager;
```
pavelprystinka, line 12: I would consider to not to include a manager in the Options class.
I would expose an event, then I would instead of assigning instance  a manager, I would subscribe on event.
The reason for this is a single responsibility principle. I like that Options class remains a data class only. When we do a composition with a manager class it now is responsible for some behaviours. Why not to keep Options class simple.

iaulakh: I thin we can use interface, instead of setting multiple event handlers.

pavelprystinka (approved):

PR #1086, opened 2024-09-18
[Calling][BugFix] Telecom manager audio selection on call join

iaulakh:
## Purpose

* On call join, if telecom manager is enabled, audio is not on speaker

- [ ] Yes
- [x] No

## Pull Request Type

- [x] Bugfix
- [ ] Feature

### DIFF
--- a/azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/redux/middleware/handler/CallingMiddlewareActionHandler.kt
+++ b/azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/redux/middleware/handler/CallingMiddlewareActionHandler.kt
@@ -901,6 +901,19 @@ internal class CallingMiddlewareActionHandlerImpl(
                     ) {
                         store.dispatch(CaptionsAction.StartRequested(localOptions.captionsOptions?.spokenLanguage ?: ""))
                     }
+
+                    // Call is only created once the call is connected (helps to fix bug on setup screen audio selection)
+                    // This helps to set correct telecom manager options on call start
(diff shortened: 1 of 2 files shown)


azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/redux/middleware/handler/CallingMiddlewareActionHandler.kt
```diff
@@ -901,6 +901,19 @@ internal class CallingMiddlewareActionHandlerImpl(
                     ) {
                         store.dispatch(CaptionsAction.StartRequested(localOptions.captionsOptions?.spokenLanguage ?: ""))
                     }
+
```
pavelprystinka, line 904: There is already a `onAudioDeviceChangeRequested` that is setting correct route.
It is better to figure out why in the Skip setup screen that is not called. Or called to late/early?
Anyway this code does not belong to here in `subscribeCallInfoModelEventUpdate`

pavelprystinka (changes requested):

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/redux/middleware/handler/CallingMiddlewareActionHandler.kt
iaulakh: Pavel lets have discussion when you have some time.

pavelprystinka (approved):

Mohtasim (approved):

PR #1096, opened 2024-09-24
[Calling][BugFix] fix random crash on accessing container

iaulakh:
## Purpose

* Fix crash where getting container cause crash
* It is random, and code was added to fix memory leak

- [ ] Yes
- [x] No

## Pull Request Type

- [x] Bugfix
- [ ] Feature

### DIFF
--- a/azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/CallCompositeActivity.kt
+++ b/azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/CallCompositeActivity.kt
@@ -225,7 +225,7 @@ internal open class CallCompositeActivity : AppCompatActivity() {
             // This code was added to avoid memory leak
             diContainerHolder.container.callCompositeActivityWeakReference = WeakReference(null)
         } catch (e: Exception) {
-            // Ignore any exception from the user handler
+            // Ignore any CallCompositeException
         }
         super.onDestroy()
     }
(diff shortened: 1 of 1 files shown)


ShaunaSong (approved):

pavelprystinka (approved):

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/CallCompositeActivity.kt
```diff
@@ -221,7 +221,12 @@ internal open class CallCompositeActivity : AppCompatActivity() {
             }
         }
 
-        diContainerHolder.container.callCompositeActivityWeakReference = WeakReference(null)
+        try {
+            // This code was added to avoid memory leak
+            diContainerHolder.container.callCompositeActivityWeakReference = WeakReference(null)
+        } catch (e: Exception) {
+            // Ignore any exception from the user handler
```
pavelprystinka, line 228: @iaulakh  could you please log an error.

ahammer: I can't see how this would throw, it's just setting a field.

I.e. it says (Ignore any exception from user handler) but there is no user handler here.


I would think this would be
```
  diContainerHolder.container.callCompositeActivityWeakReference.clear()
```

But really, that doesn't make sense, because it's a weak reference and it should auto-clear when there is no references to it. So if the WeakReference isn't clearing another reference is keeping it alive. (I.e. the Context is still leaking probably).

ahammer: can you share the stacktrace for the crash?

pavelprystinka (approved):

ahammer (approved):

PR #1125, opened 2024-12-09
[Calling][Android] Feature RTT

pavelprystinka:
## Purpose

* Feature RTT

- [ ] Yes
- [X] No

## Pull Request Type

- [ ] Bugfix
- [X] Feature

## Pull Request Checklist

- [ ] Public API changes
- [ ] UI Changes
- [ ] Tests
  - [ ] Unit Tests Included

## How to Test

* Open...
* Click on...

## What to Check
* ...


azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/CallCompositeActivity.kt
```diff
@@ -100,6 +103,13 @@ internal open class CallCompositeActivity : AppCompatActivity() {
             return
         }
 
+        store.dispatch(DeviceConfigurationAction.IsTableChanged(isTablet(this)))
```
ahammer, line 106: Action, so should be verbiage (I.e. DeviceConfigruation.ToggleTabletMode. I.e. where "Toggle" is the action. 

Also Table->Tablet

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/presentation/fragment/calling/CallingFragment.kt
```diff
@@ -275,19 +306,45 @@ internal class CallingFragment :
...
+                val captionsWrapperLayout =
+                    captionsWrapper.layoutParams as ConstraintLayout.LayoutParams
+                captionsWrapperLayout.matchConstraintPercentWidth =
+                    if (isLandScape) 0.33f else 0.45f
+                captionsWrapper.layoutParams = captionsWrapperLayout
+            }
+
+            viewLifecycleOwner.lifecycleScope.launch {
+                viewModel.isCaptionsVisibleFlow.collect {
+                    val height = if (it) 150 else 0
```
ahammer, line 329: Magic #'s to Constants, or preferable, dimens.xml or resource files that can be configuration selected better.

jamchengms (dismissed):

azure-communication-ui/build.gradle
```diff
@@ -46,7 +46,7 @@ buildscript {
             // and enable the beta features
 
             // Latest GA version of the SDK
-            azure_calling_sdk_version_ga = '2.12.0'
+            azure_calling_sdk_version_ga = '2.13.0-beta.1'
 
             // Newest version of the SDK, can be beta
             azure_calling_sdk_version_beta = '2.12.0-beta.1'
```
ahammer, line 52: I know there are no flags, but once this is merged, we are hard-committed to 2.13 GA (which might be delayed due to TDL bugs).

Flags, and using the _beta field would help in mitigating this risk, but it's a lot to ask at this point. 

If something comes up and we need adjust path to GA without RTT (for any reason), we'll have to strategically flag to disable when going back to 2.12, or hotfix on top of the last release. It's worth acknowledging this risk before merging. Best case it's a non-issue.

pavelprystinka: We did discuss it: as there is no other feature in development that may go sooner, I did not add a feature flag. Otherwise, it would be much harder to do all those refactorings and renamings.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/service/CallingService.kt
```diff
@@ -3,18 +3,20 @@
 
 package com.azure.android.communication.ui.calling.service
 
+/*  <CALL_START_TIME> */
```
ahammer, line 6: Broken flag for call start time. (covers all imports).

Recommend removing, as if the feature gets disabled, the error will propagate anyways, and can then be fixed in the main class file (i.e. by just stubbing out it's impl, but allowing the import to remain without flags).

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/service/CallingService.kt
```diff
@@ -188,16 +184,13 @@ internal class CallingService(
...
-            callingSdk.getRttSharedFlow().collect {
-                rttStateSharedFlow.emit(it)
-            }
-        }
-        </RTT_POC> */
         return callingSdk.startCall(cameraState, audioState)
     }
 
+    fun sendRttMessage(message: String, isFinalized: Boolean) {
+        callingSdk.sendRttMessage(message, isFinalized)
```
ahammer, line 191: you may want to trim here, as sendMessage can fail if string is too long

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/service/sdk/CallingSDK.kt
```diff
@@ -3,6 +3,7 @@
 
 package com.azure.android.communication.ui.calling.service.sdk
 
+/*  <CALL_START_TIME> */
```
ahammer, line 6: Same, remove any flags in import sections. If CALL_START_TIME needs disabling, we can fix the breakages when they happen by adjusting other flags.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/service/CallingService.kt
pavelprystinka: It is trimmed on the UI layer. I think we should not quietly hide the fact that callingSdk.sendRttMessage("too long message", isFinalized) is actually sends a shorter message.

azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/service/sdk/CallingSDKEventHandler.kt
```diff
@@ -3,18 +3,14 @@
 
 package com.azure.android.communication.ui.calling.service.sdk
 
+/*  <CALL_START_TIME> */
```
ahammer, line 6: Mangled flags

azure-communication-ui/build.gradle
ahammer: We can just do it surgically if it comes down to it, it'd be way less work than doing the entire feature. I'm not expecting on this PR, just bringing up the risk.

ahammer (approved):

ShaunaSong (approved):

